PR #7, opened 2023-02-08
fixed all clang-tidy issues

RisingLiberty:


source/2_platform/rex_windows/src/gui_application.cpp
```diff
@@ -64,28 +63,17 @@ namespace rex
...
         window->update();
-        world.update();
 
         is_running = !is_marked_for_destroy;
 
         //
         // Cap framerate to "max_fps".
         // Safe resources of the machine we are running on.
         //
-        const rsl::chrono::milliseconds actual_time(static_cast<int64>(rsl::lrint(1000.0f / static_cast<f32>(world.get_frames_per_seconds().get()))));
```
Dyronix, line 80: Why is this removed, we do not want to fly through the app. We should cap it on a specific framerate IMO.

source/2_platform/rex_windows/src/win_window.cpp
```diff
@@ -67,10 +67,25 @@ namespace rex
...
 
+      // clang-format off
       m_hwnd = WIN_CALL(
-          (HWND)CreateWindowA(description.title, description.title, WS_OVERLAPPEDWINDOW, x == 0 ? screen_mid_x - half_x : x, y == 0 ? screen_mid_y - half_y : y, rc.right - rc.left, rc.bottom - rc.top, nullptr, nullptr, (HINSTANCE)hInstance, this));
-
-      if(m_hwnd == NULL)
+          static_cast<HWND>(CreateWindowA(
+            description.title, 
+            description.title, 
+            WS_OVERLAPPEDWINDOW, x == 0
```
Dyronix, line 75: I always indent there, this is not readable ...

source/2_platform/rex_windows/src/win_window.cpp
```diff
@@ -67,10 +67,25 @@ namespace rex
...
+             ? screen_mid_x - half_x 
+             : x,
+             y == 0
```
Dyronix, line 78: I always indent there, this is not readable ...

source/2_platform/rex_windows/src/win_window_class.cpp
```diff
@@ -15,15 +15,15 @@ namespace rex
...
+          , hinstance(nullptr)
           , registered(false)
       {
       }
 
       bool create(HInstance hInstance, WindowProcedureFunc wndProc, const char8* title)
       {
         name      = title;
-        hinstance = (hInstance == nullptr) ? static_cast<HInstance>(GetModuleHandleA(NULL)) : hInstance;
+        hinstance = (hInstance == nullptr) ? static_cast<HInstance>(GetModuleHandleA(nullptr)) : hInstance;
```
Dyronix, line 26: When using the WIN32 API I rather use NULL instead of nullptr. Although I am not consistent but that is another review comment.

Dyronix (changes requested):

source/2_platform/rex_windows/src/gui_application.cpp
RisingLiberty: there's no world anymore, we'll put it back when we figure out fps

source/2_platform/rex_windows/src/win_window_class.cpp
RisingLiberty: clang-tidy complain when using NULL, which is why I changed it

PR #16, opened 2023-02-10
Develop/memory manager

RisingLiberty:


source/1_engine/rex_engine/include/rex_engine/core_application.h
```diff
@@ -7,16 +7,16 @@ namespace rex
 {
   struct FrameInfo;
 
-  class IApplication
+  class RexApplication
```
Dyronix, line 10: Not so sure if I would call this RexApplication ... this does not represent what it actually is IMO. It's a BaseApplication, so maybe just call it that. I know calling it a RexApplication might be cool but it could become a bit confusing in the inheritance tree where this is located

source/1_engine/rex_engine/include/rex_engine/memory/memory_manager.h
```diff
@@ -0,0 +1,25 @@
...
+  public:
+    MemoryManager();
+
+    void initialize(rsl::memory_size maxMemUsage);
+
+    void track_alloc(u64 size);
+    void track_dealloc(u64 size);
+
+  private:
+    rsl::high_water_mark<s64> m_mem_usage; // current memory usage
```
Dyronix, line 20: I would create a using for this to make the name of this object a little more clear. I have no idea type of object this is without looking it up.

To me a water mark is something that is put over an image, this has nothing to do with it.

source/1_engine/rex_engine/include/rex_engine/types.h
```diff
@@ -22,12 +22,12 @@ using ulong = unsigned long;
 
 // allow single threaded platforms to avoid use of atomic
 #if REX_WEB || REX_WINDOWS
-  #define REX_SINGLE_THREADED 1
+  #define REX_SINGLEHREADED 1
```
Dyronix, line 25: Typo
REX_SINGLE - T - HREADED

Dyronix (changes requested):

source/1_engine/rex_engine/include/rex_engine/core_application.h
RisingLiberty: ok

source/1_engine/rex_engine/include/rex_engine/memory/memory_manager.h
RisingLiberty: then a using won't make a different, you're suggesting a type name change, which is fine, but I don't know what else to call it.
it's a high water mark, like people do with rivers to track how high up it has come

source/1_engine/rex_engine/include/rex_engine/types.h
RisingLiberty: ok

source/1_engine/rex_engine/include/rex_engine/memory/memory_manager.h
RisingLiberty: A high water mark is a point that represents the maximum rise of a body of water over land.
https://en.wikipedia.org/wiki/High_water_mark

PR #24, opened 2023-02-16
Development/renderer initialization review

Dyronix:
Review comments applied ( excluded clang tidy )


source/2_platform/rex_opengl/include/rex_opengl/resources/buffer.h
```diff
@@ -16,8 +16,13 @@ namespace rex
     {
     public:
         Buffer(u32 target);
+        Buffer(const Buffer&) = delete;
+        Buffer(Buffer&&) = default;
```
RisingLiberty, line 20: the default move ctor will copy the resource id without clearing the one of the input.
I don't think this is intended, please define a custom implementation that does this.

source/2_platform/rex_opengl/include/rex_opengl/resources/buffer.h
```diff
@@ -16,8 +16,13 @@ namespace rex
...
         ~Buffer() override;
 
+        Buffer& operator=(const Buffer&) = delete;
+        Buffer& operator=(Buffer&&) = default;
```
RisingLiberty, line 24: same as for move ctor, but this one should act as if the destructor of *this is called, and then the move ctor is called

source/2_platform/rex_opengl/include/rex_opengl/resources/buffer_layout.h
```diff
@@ -33,35 +37,35 @@ namespace rex
...
-        inline ConstReverseBufferLayoutIt     crend() const { return m_elements.crend(); }
-
     public:
-        BufferLayout(const BufferLayout& other);
-        BufferLayout(BufferLayout&& other) noexcept;
+        BufferLayout(const BufferElements& elements);
 
-        BufferLayout& operator=(const BufferLayout& other);
-        BufferLayout& operator=(BufferLayout&& other) noexcept;
+        BufferLayout(const BufferLayout&) = default;
```
RisingLiberty, line 43: if they're all default, they can be deleted

source/2_platform/rex_opengl/src/resources/buffer_layout.cpp
```diff
@@ -4,170 +4,132 @@
...
+
+  }
+  //-----------------------------------------------------------------------
+  BufferLayout& BufferLayoutBuilder::add_buffer_element(DataType inType, ShouldNormalize inShouldNormalize /*= ShouldNormalize::no*/)
+  {
+    BufferElements new_buffer_elements;
+
+    new_buffer_elements.reserve(m_layout.size() + 1);
+    new_buffer_elements.insert(rsl::end(new_buffer_elements), rsl::begin(m_layout.get_buffer_elements()), rsl::end(m_layout.get_buffer_elements()));
+    new_buffer_elements.emplace_back(BufferElement(inType, buffer_layout::get_data_type_component_count(inType), inShouldNormalize));
```
RisingLiberty, line 84: with emplace_back, you can drop the BufferElement constructor call, just pass in the arguments you'd pass to the constructor

RisingLiberty (changes requested):

source/2_platform/rex_opengl/include/rex_opengl/resources/buffer.h
Dyronix: There should not be any destruction here, there should ony be a move of the resource id
If I call glDeleteBuffer on the given id the buffer that we moved to will also be invalid

source/1_engine/rex_engine/include/rex_engine/defines.h
```diff
@@ -24,10 +24,10 @@
 
 //-------------------------------------------------------------------------
 // Assembly instruction to break execution.
-#if defined _MSC_VER
+#if defined __clang__
```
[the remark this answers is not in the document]
Dyronix: If I use your macros it does not work

source/2_platform/rex_opengl/src/opengl_renderer.cpp
```diff
@@ -102,14 +119,15 @@ namespace rex
...
-        // unused on this platform
+        // Nothing to implement
       }
 
-      // clears
       void clear()
       {
         ClearState& cs = s_clear_state;
 
+        REX_ASSERT(cs.flags);
```
RisingLiberty, line 129: there is no ASSERT message here

RisingLiberty (changes requested):

PR #41, opened 2023-02-23
Development/rex externals

Dyronix:


build/config/settings.json
```diff
@@ -8,5 +8,5 @@
   "coverage_folder": "tests/coverage",
   "asan_folder": "tests/coverage",
   "ubsan_folder": "tests/ubsan",
-  "ninja_launcher": "build/scripts/rexpy/rexpy/ninja_launcher.py"
+  "ninja_launcher": "build/scripts/rex-py/rexpy/ninja_launcher.py"
```
RisingLiberty, line 11: naming guideline I follow is that folders don't have '-' in there.
where did this come from?

_setup.py
```diff
@@ -8,29 +8,23 @@
...
+  p.communicate()
 
-  # first let's make sure we're in the root directory
-  # we don't have access to rexpy yet, we'll have to manually
-  # check if we're in the root.
+def __main():
+  parser = argparse.ArgumentParser()
+  args, unknown = parser.parse_known_args()
 
   # let's assume if "build" and "source"are found, we're in the root
```
RisingLiberty, line 29: update this comment

_setup.py
```diff
@@ -8,29 +8,23 @@
...
-  return False
+def __get_script_path():
+    return os.path.dirname(os.path.realpath(__file__))
 
-def install_rexpy(forceInstall):
-  # first set our working directory to rexpy.
-  # this way the package will be installed in {root}/build/scripts/rexpy
-  # this keeps the root clean.
+def __install_externals():
+  install_externals_path = os.path.join(__get_script_path(), os.path.join("build", os.path.join("scripts", "install_externals.py")))
```
RisingLiberty, line 21: this path should be wrapped in quotes

_setup.py
```diff
@@ -40,39 +34,16 @@ def install_rexpy(forceInstall):
...
-  parser.add_argument("-update_rexpy", help="update rexpy, this is basically a reinstall", action="store_true")
-
-  args, unknown = parser.parse_known_args()
-
-  install_rexpy(args.update_rexpy)
-
-  # now that rexpy is installed, we can safely call the rest of the code
+  __install_externals()
 
+  # now that rexpy and rexstd is installed, we can safely call the rest of the code
```
RisingLiberty, line 39: update this comment

build/scripts/install_externals.py
```diff
@@ -0,0 +1,207 @@
...
+class Host(Enum):
+    UNKNOWN = 0
+    GITLAB = 1
+    GITHUB = 2
+
+def __get_script_path():
+    return os.path.dirname(os.path.realpath(__file__))
+def __get_root_path():
+    root = __get_script_path()
+    root = os.path.join(root, os.path.join("..", ".."))
```
RisingLiberty, line 17: look for the build folder in parent, don't use relative paths

build/scripts/install_externals.py
```diff
@@ -0,0 +1,207 @@
...
+    if "gitlab" in path:
+        return Host.GITLAB
+    elif "github" in path:
+        return Host.GITHUB
+    
+    print("Unknown host!")
+    return Host.UNKNOWN
+
+def __build_gitlab_path(baseUrl, name, tag):
+    url = os.path.join(baseUrl, "-")
```
RisingLiberty, line 31: write a comment what the format of gitlab looks like

build/scripts/install_externals.py
```diff
@@ -0,0 +1,207 @@
...
+    url = os.path.join(url, "archive")
+    url = os.path.join(url, tag)
+    url = os.path.join(url, name+"-"+tag+".zip")
+    url = url.replace("\\", "/")
+    return url
+def __build_github_path(baseUrl, tag):
+    url = os.path.join(baseUrl, "archive")
```
RisingLiberty, line 38: write a comment what the format of github looks like

build/scripts/install_externals.py
```diff
@@ -0,0 +1,207 @@
...
+    url = os.path.join(url, "refs")
+    url = os.path.join(url, "tags")
+    url = os.path.join(url, tag+".zip")
+    url = url.replace("\\", "/")
+    return url
+def __build_host_path(baseUrl, name, tag):
+    host = __get_host(baseUrl)
+    match host:
```
RisingLiberty, line 46: match is only supported in python 3.10, a lot of people won't be able to run this

build/scripts/install_externals.py
```diff
@@ -0,0 +1,207 @@
...
+    root = __get_root_path()
+
+    external_url = external["url"]
+    external_name = external["name"]
+    external_tag = external["tag"]
+    external_store = external["storage"]
+    external_store = external_store.replace("~", root)
+
+    # if the external is already present we need to check if we need to redownload anything
+    should_download = False
```
RisingLiberty, line 145: put this in a function

build/scripts/install_externals.py
```diff
@@ -0,0 +1,207 @@
...
+    externals_dir = os.path.join(external_store, external_name)
+    if os.path.exists(externals_dir):
+        print("External found: " + external_name + " validating version ...")
```
RisingLiberty, line 148: also print where you found it

build/scripts/install_externals.py
```diff
@@ -0,0 +1,207 @@
...
+        os.chdir(externals_dir)
```
RisingLiberty, line 149: I don't think you have to change the directory, keep using the absolute paths, it avoids a missing error where you put the cwd back to what it was

build/scripts/install_externals.py
```diff
@@ -0,0 +1,207 @@
...
+            version_data = __load_json(version_file)           
+            if version_data == None:
+                print("Invalid version data found, redownloading external: " + external_name)
+                should_download = True              
+            if not version_data["tag"] == external_tag:
+                should_download = True
+            else:
+                print("External: " + external_name + " is up to date (" + external_name + " " + external_tag + ")")
+        else:
+            should_download = True
```
RisingLiberty, line 161: print that you didn't find a version file

build/scripts/install_externals.py
```diff
@@ -0,0 +1,207 @@
+import os
```
RisingLiberty, line 1: this entire script should be moved to rexpy, now I know this script also downloads rexpy so I'm not sure how we want to solve this.
it's possible a hardcoded rexpy url to download works and we donwload it uncompressed and then call into rexpy to download the rest

_setup.py
```diff
@@ -40,39 +34,16 @@ def install_rexpy(forceInstall):
...
-if __name__ == "__main__":
-  parser = argparse.ArgumentParser()
-  parser.add_argument("-update_rexpy", help="update rexpy, this is basically a reinstall", action="store_true")
-
-  args, unknown = parser.parse_known_args()
-
-  install_rexpy(args.update_rexpy)
-
-  # now that rexpy is installed, we can safely call the rest of the code
+  __install_externals()
```
RisingLiberty, line 37: pip needs to rerun on rexpy if it's downloaded to addd it to the python libraries

RisingLiberty (changes requested):

build/config/settings.json
Dyronix: RexStd was called rex_std, RexPy was called rexpy and glad & glm where called rex-glad and rex-glm I just picked one

Dyronix: If files are not allowed to have no spaces I think folders should follow a similar guideline and I quite like the "-" between rex and the name of the library

RisingLiberty: spaces are not recommened, I stuck with '_' because the source also has underscores

Dyronix: I will change it to an "_"

build/scripts/install_externals.py
Dyronix: We will use rexpy for this which will be installed using pip

build/scripts/install_externals.py
Dyronix: We will use rexpy for this which will be installed using pip

PR #48, opened 2023-03-04
Compiler defines added | Enabled assert in non shipping builds

RisingLiberty:


source/1_engine/rex_engine/include/rex_engine/cmd_line_args.h
```diff
@@ -1,6 +1,6 @@
 #pragma once
 
-#include "rex_std/string_view.h"
+#include "rex_std/internal/string_view/basic_string_view.h"
```
Dyronix, line 3: Can you replace the internal include to the regular one within rex please?
This comment applies to all the includes within rex

source/2_platform/rex_opengl/src/opengl_renderer.cpp
```diff
@@ -18,6 +18,7 @@ namespace rex
   //-------------------------------------------------------------------------
   enum class ClearBits
   {
+    None               = 0,
```
Dyronix, line 21: I would opt to make this a negative value, OpenGL might have a reserved place for 0 maybe not in this case but just to be consistent. Negative values are most certainly invalid though.

Dyronix (changes requested):

source/1_engine/rex_engine/include/rex_engine/diagnostics/win/hr_call.h
```diff
@@ -13,7 +13,7 @@
 #pragma once
 
 #include "rex_std/bonus/string/stack_string.h"
-#include "rex_std/string_view.h"
+#include "rex_std/internal/string_view/basic_string_view.h"
```
Dyronix, line 16: Can you replace the internal include to the regular one within rex please?
This comment applies to all the includes within rex

Dyronix (changes requested):

source/1_engine/rex_engine/include/rex_engine/diagnostics/win/win_call.h
```diff
@@ -14,7 +14,7 @@
 
 #include "rex_std/bonus/string/stack_string.h"
 #include "rex_std/bonus/types.h"
-#include "rex_std/string_view.h"
+#include "rex_std/internal/string_view/basic_string_view.h"
```
Dyronix, line 17: Can you replace the internal include to the regular one within rex please?
This comment applies to all the includes within rex

Dyronix (changes requested):

PR #49, opened 2023-03-04
buffer layout optimization

RisingLiberty:
optimizing buffer layout builder to avoid copying of vectors every time a new buffer element is added

### DIFF
--- a/source/2_platform/rex_opengl/include/rex_opengl/resources/buffer_layout.h
+++ b/source/2_platform/rex_opengl/include/rex_opengl/resources/buffer_layout.h
@@ -108,10 +108,11 @@ namespace rex
   public:
     BufferLayoutBuilder();
 
-    BufferLayout& add_buffer_element(DataType::Value inType, ShouldNormalize inShouldNormalize = ShouldNormalize::no);
-    BufferLayout& build();
+    void add_buffer_element(DataType::Value inType, ShouldNormalize inShouldNormalize = ShouldNormalize::no);
+    void clear();
+    BufferLayout build();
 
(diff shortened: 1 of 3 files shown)


source/2_platform/rex_opengl/include/rex_opengl/resources/buffer_layout.h
```diff
@@ -108,10 +108,11 @@ namespace rex
   public:
     BufferLayoutBuilder();
 
-    BufferLayout& add_buffer_element(DataType::Value inType, ShouldNormalize inShouldNormalize = ShouldNormalize::no);
-    BufferLayout& build();
+    void add_buffer_element(DataType::Value inType, ShouldNormalize inShouldNormalize = ShouldNormalize::no);
```
Dyronix, line 111: This is not a builder pattern anymore.
The reason this was created was to make the API when using this class a lot more readable I would opt to keep it at that, 
Did you profile this, before you added the optimization?

Dyronix (changes requested):

source/2_platform/rex_opengl/include/rex_opengl/resources/buffer_layout.h
Dyronix: We discussed this and we agreed to not call this a builder anymore.
Convert this class to a wrapper around a vector which only exposes the required functionality
Naming can be anything as long as it starts with the prefix BufferLayout ( eg: BufferLayoutGenerator, BufferLayoutCreator, BufferLayoutAssembler, ... )

PR #72, opened 2023-04-16
Development/directx

Dyronix:


RisingLiberty:
please fix the issues with CI

Dyronix:
We are only allowed to merge this when "development/assert_system" has been fixed and review remarks are applied

source/2_platform/rex_directx/rex_directx.sharpmake.cs
```diff
@@ -0,0 +1,31 @@
...
+  }
+
+  public override void Configure(RexConfiguration conf, RexTarget target)
+  {
+    base.Configure(conf, target);
+
+    conf.Output = Configuration.OutputType.Lib;
+
+    conf.AddPublicDependency<RexStdExtra>(target);
+    conf.AddPublicDependency<RexRendererCore>(target);
```
RisingLiberty, line 28: you need to enable the clang-tidy header dependency here as well

source/2_platform/rex_directx/src/directx_context.cpp
```diff
@@ -0,0 +1,21 @@
...
+namespace rex
+{
+  namespace context
+  {
+    namespace backend
+    {
+      //-------------------------------------------------------------------------
+      bool create(void* /*unused*/)
+      {
+        return true;
```
RisingLiberty, line 15: I assume this will be implemented in the future?

source/2_platform/rex_directx/src/directx_context.cpp
```diff
@@ -0,0 +1,21 @@
+#include "rex_directx/log.h"
+#include "rex_renderer_core/context.h"
+
+// NOLINTBEGIN(modernize-use-nullptr,-warnings-as-errors)
```
RisingLiberty, line 4: why is this disabled?

source/2_platform/rex_directx/src/directx_feature_level.cpp
```diff
@@ -0,0 +1,49 @@
...
+#include "rex_directx/dxgi/util.h"
+#include "rex_engine/diagnostics/assert.h"
+#include "rex_std/array.h"
+
+namespace rex
+{
+  const rsl::array g_expected_feature_levels = {D3D_FEATURE_LEVEL::D3D_FEATURE_LEVEL_12_0, D3D_FEATURE_LEVEL::D3D_FEATURE_LEVEL_12_1};
+
+  //-------------------------------------------------------------------------
+  rsl::string feature_level_name(const D3D_FEATURE_LEVEL& level)
```
RisingLiberty, line 13: this can return a stack string and can use the enum reflection to avoid writing the function in the first place

source/2_platform/rex_directx/include/rex_directx/directx_feature_level.h
```diff
@@ -0,0 +1,19 @@
...
+#include "rex_engine/types.h"
+#include "rex_std/string.h"
+
+struct IDXGIAdapter;
+
+namespace rex
+{
+  struct FeatureLevelInfo
+  {
+    s32 level;
```
RisingLiberty, line 12: this seams to be storing the int version of the enum value.
I don't think this is necessary and we have enum reflection to get the name.
so I think this struct can be removed all together

source/2_platform/rex_directx/src/directx_feature_level.cpp
```diff
@@ -0,0 +1,49 @@
+#include "rex_directx/directx_feature_level.h"
```
RisingLiberty, line 1: see my comments in the header regarding this functionality, I don't think it's needed

source/2_platform/rex_directx/src/directx_feature_shader_model.cpp
```diff
@@ -0,0 +1,76 @@
+#include "rex_directx/directx_feature_shader_model.h"
+
+#include "rex_directx/directx_util.h"
+
+namespace rex
+{
+  //-------------------------------------------------------------------------
+  rsl::string shader_model_name(s32 shaderModel)
```
RisingLiberty, line 8: can use stack string for this.
can even use enum reflection to avoid this function to begin with

source/2_platform/rex_directx/src/directx_feature_shader_model.cpp
```diff
@@ -0,0 +1,76 @@
...
+
+    return SUCCEEDED(hr);
+  }
+
+  //-------------------------------------------------------------------------
+  ShaderModelInfo query_shader_model_version(ID3D12Device* device)
+  {
+    D3D12_FEATURE_DATA_SHADER_MODEL shader_model = {};
+
+#if defined(NTDDI_WIN10_VB) && (NTDDI_VERSION >= NTDDI_WIN10_VB)
```
RisingLiberty, line 39: I believe the expected shader model should be constant and not based on what's available.
this will return different results depending on which machine it's compiled

source/2_platform/rex_directx/include/rex_directx/directx_feature_shader_model.h
```diff
@@ -0,0 +1,21 @@
...
+
+#include "rex_engine/types.h"
+#include "rex_std/string.h"
+
+struct ID3D12Device;
+
+namespace rex
+{
+  struct ShaderModelInfo
+  {
```
RisingLiberty, line 11: the same comment applies as the directx feature level.
I don't think this struct is needed, we have enum reflection to get the name and the int represenation hides what it's supposed to be and will only add complexity

source/2_platform/rex_directx/src/directx_feature_shader_model.cpp
```diff
@@ -0,0 +1,76 @@
+#include "rex_directx/directx_feature_shader_model.h"
+
+#include "rex_directx/directx_util.h"
+
+namespace rex
+{
+  //-------------------------------------------------------------------------
+  rsl::string shader_model_name(s32 shaderModel)
+  {
```
RisingLiberty, line 9: see my comment in the header regarding the necessesity of this functionality

source/2_platform/rex_directx/src/directx_renderer.cpp
```diff
@@ -0,0 +1,122 @@
...
+
+      DirectXContext g_ctx; // NOLINT(fuchsia-statically-constructed-objects, cppcoreguidelines-avoid-non-const-global-variables)
+
+      //-------------------------------------------------------------------------
+      bool initialize()
+      {
+        rsl::unique_ptr<dxgi::Factory> factory = dxgi::Factory::create();
+
+        dxgi::AdapterManager adapter_manager(factory.get(), HighestVramGpuScorer());
+        const dxgi::Adapter* adapter = static_cast<const dxgi::Adapter*>(adapter_manager.selected()); // NNOLINT(cppcoreguidelines-pro-type-static-cast-downcast)
```
RisingLiberty, line 60: please look at the comments I've given regarding the Gpu design.
I think as soon as you have to down cast to a type, there's something wrong with the code design.
in this case the down casting can be prevented

source/2_platform/rex_directx/src/directx_renderer.cpp
```diff
@@ -0,0 +1,122 @@
...
+
+        const FeatureLevelInfo feature_level_info = query_feature_level(adapter->c_ptr());
+
+        if(FAILED(D3D12CreateDevice(adapter->c_ptr(), static_cast<D3D_FEATURE_LEVEL>(feature_level_info.level), IID_PPV_ARGS(&g_ctx.device))))
+        {
+          if(adapter_manager.load_software_adapter(factory.get()))
+          {
+            REX_ERROR(LogDirectX, "Could not create software adapter");
```
RisingLiberty, line 68: load_software_adapter returns true on success, not on failure

RisingLiberty: do we want to support software adaptors? I see little point in doing so

source/2_platform/rex_directx/src/directx_renderer.cpp
```diff
@@ -0,0 +1,122 @@
...
+          {
+            const dxgi::Adapter* software_adapter = static_cast<const dxgi::Adapter*>(adapter_manager.software()); // NNOLINT(cppcoreguidelines-pro-type-static-cast-downcast)
+            if(FAILED(D3D12CreateDevice(software_adapter->c_ptr(), static_cast<D3D_FEATURE_LEVEL>(feature_level_info.level), IID_PPV_ARGS(&g_ctx.device))))
+            {
+              REX_ERROR(LogDirectX, "Failed to create DX12 Device");
+              return false;
+            }
+          }
+        }
+        else
```
RisingLiberty, line 82: else statement isn't really needed here

source/2_platform/rex_directx/src/directx_renderer.cpp
```diff
@@ -0,0 +1,122 @@
...
+      }
+
+      //-------------------------------------------------------------------------
+      void end_frame()
+      {
+        // Nothing to implement
+      }
+
+      //-------------------------------------------------------------------------
+      void clear() {}
```
RisingLiberty, line 116: nothing to implement statement? 
I know it'll be implement in the future, but just to be consistent with the rest

source/2_platform/rex_directx/src/directx_renderer.cpp
```diff
@@ -0,0 +1,122 @@
...
+
+      //-------------------------------------------------------------------------
+      void present() {}
```
RisingLiberty, line 119: nothing to implement statement? 
I know it'll be implement in the future, but just to be consistent with the rest

source/2_platform/rex_directx/src/directx_util.cpp
```diff
@@ -0,0 +1 @@
+#include "rex_directx/directx_util.h"
```
RisingLiberty, line 1: please don't add empty files

source/2_platform/rex_directx/src/dxgi/adapter.cpp
```diff
@@ -0,0 +1,130 @@
...
+    {
+      case Vendors::AMD: return rsl::string("AMD");
+      case Vendors::NVIDIA: return rsl::string("NVIDIA");
+      case Vendors::INTEL: return rsl::string("INTEL");
+      default: return rsl::string("Unknown Vendor");
+    }
+  }
+
+  ////-------------------------------------------------------------------------
+  // rsl::string to_multibyte(const tchar* wideCharacterBuffer, count_t size)
```
RisingLiberty, line 39: shouldn't this get removed?

source/2_platform/rex_directx/src/dxgi/adapter.cpp
```diff
@@ -0,0 +1,130 @@
...
+  enum class Vendors
+  {
+    UNKNOWN = 0,
+    AMD     = 0x1002,
+    NVIDIA  = 0x10DE,
+    INTEL   = 0x163C
+  };
+
+  //-------------------------------------------------------------------------
+  rsl::string vendor_to_string(s32 vendor)
```
RisingLiberty, line 27: use enum reflection for this

source/2_platform/rex_directx/src/dxgi/adapter.cpp
```diff
@@ -0,0 +1,130 @@
...
+  //     // size_t converted_chars => The amount of converted characters.
+  //     // 0 terminate the string afterwards.
+  //     count_t converted_chars = rsl::wcstombs(buffer.data(), wideCharacterBuffer, size);
+  //     buffer.resize(converted_chars);
+  //
+  //     return buffer;
+  // }
+
+  //-------------------------------------------------------------------------
+  rsl::string to_multibyte(const tchar* wideCharacterBuffer, count_t size)
```
RisingLiberty, line 53: can use a stack string here

source/2_platform/rex_directx/src/dxgi/adapter.cpp
```diff
@@ -0,0 +1,130 @@
...
+    return desc;
+  }
+} // namespace
+
+namespace rex
+{
+  namespace dxgi
+  {
+    //-------------------------------------------------------------------------
+    Adapter::Adapter(wrl::com_ptr<IDXGIAdapter>&& adapter, uint32 version)
```
RisingLiberty, line 124: please see my comment in the header regarding this class

source/2_platform/rex_directx/include/rex_directx/dxgi/adapter.h
```diff
@@ -0,0 +1,23 @@
...
+#include <string>
+
+struct IDXGIAdapter;
+
+namespace rex
+{
+  namespace dxgi
+  {
+    class Adapter : public Gpu, public dxgi::ComObject<IDXGIAdapter> // NOLINT(fuchsia-multiple-inheritance)
+    {
```
RisingLiberty, line 18: I realise that you use double inheritance here and want the functionality of the com object as well, I'm just wondering if there's a better, less complex way of designing it

source/2_platform/rex_directx/src/dxgi/adapter_manager.cpp
```diff
@@ -0,0 +1,154 @@
...
+
+namespace rex
+{
+  namespace dxgi
+  {
+    //-------------------------------------------------------------------------
+    AdapterManager::AdapterManager(const Factory* factory, const GpuScorer& scorer)
+        : m_selected_adapter(nullptr)
+    {
+      load_adapters(factory);
```
RisingLiberty, line 88: shouldn't this assert that you can always load at least 1 (load_adapters returns true if any adapters are loaded)

source/2_platform/rex_directx/src/dxgi/comobject.cpp
```diff
@@ -0,0 +1 @@
+#include "rex_directx/dxgi/comobject.h"
```
RisingLiberty, line 1: please don't add empty cpp files

source/2_platform/rex_directx/src/dxgi/util.cpp
```diff
@@ -0,0 +1 @@
+#include "rex_directx/dxgi/util.h"
```
RisingLiberty, line 1: please don't add empty cpp files

source/2_platform/rex_directx/src/wrl/comobject.cpp
```diff
@@ -0,0 +1 @@
+#include "rex_directx/wrl/comobject.h"
```
RisingLiberty, line 1: please don't add empty cpp files

source/2_platform/rex_directx/src/wrl/wrl_types.cpp
```diff
@@ -0,0 +1 @@
+#include "rex_directx/wrl/wrl_types.h"
```
RisingLiberty, line 1: please don't add empty cpp files

source/2_platform/rex_windows/src/win_directx_context.cpp
```diff
@@ -0,0 +1,28 @@
+#ifdef REX_API_DIRECTX12
```
RisingLiberty, line 1: with lateest sharpmake version, the ifdef guard can be removed

source/2_platform/rex_windows/src/win_directx_context.cpp
```diff
@@ -0,0 +1,28 @@
...
+      bool pre_setup(void* /*userData*/)
+      {
+        return true;
+      }
+      bool post_setup(void* /*userData*/)
+      {
+        return true;
+      }
+
+      void rollback(void* /*userData*/) {}
```
RisingLiberty, line 24: forgot nothing to implement statement?

source/2_platform/rex_windows/src/win_opengl_context.cpp
```diff
@@ -1,10 +1,11 @@
-#include "rex_engine/types.h"
-#include "rex_renderer_core/context.h"
-#include "rex_windows/log.h"
+#ifdef REX_API_OPENGL
```
RisingLiberty, line 1: with latest sharpmake the ifdef guard can be removed here

RisingLiberty (changes requested):

source/2_platform/rex_directx/src/dxgi/adapter.cpp
Dyronix: Depends are you going to put it into rsl or not? 
Cuz that's the reason why it is there

RisingLiberty: the C standard has support for this afaik wctomb?

Dyronix: Yes and we don't cover some of those functions in RSL?

Dyronix: I use wcstombs_s as we speak

source/2_platform/rex_directx/src/directx_context.cpp
Dyronix: Yes it will, once this PR is in main

source/2_platform/rex_directx/src/directx_context.cpp
Dyronix: copy paste error

source/2_platform/rex_directx/src/directx_feature_shader_model.cpp
Dyronix: Yes currently there are no settings available so I will keep it like this, I'll add an issue that explain we need some for of engine settings which will store the shader model and the fallback model

source/2_platform/rex_directx/src/directx_renderer.cpp
```diff
@@ -0,0 +1,137 @@
...
+namespace rex
+{
+  namespace renderer
+  {
+    namespace internal
+    {
+      //-------------------------------------------------------------------------
+      const Gpu* highest_scoring_gpu(const rsl::vector<rsl::unique_ptr<Gpu>>& gpus)
+      {
+        auto it = std::max_element(gpus.cbegin(), gpus.cend(),
```
RisingLiberty, line 21: should we assert that the gpus array length should be higher than 0?

source/2_platform/rex_directx/src/dxgi/adapter.cpp
```diff
@@ -0,0 +1,118 @@
...
+    auto result            = wcstombs_s(&converted_chars, buffer.data(), size, wideCharacterBuffer, size);
+    if(result != 0)
+    {
+      REX_ERROR(LogDirectX, "Failed to convert wide character string to multi byte character string.");
+      return rsl::string("Invalid String");
+    }
+
+    buffer.resize(converted_chars);
+
+    return rsl::string(buffer.data()); // NOLINT(readability-redundant-string-cstr)
```
RisingLiberty, line 58: provide the length of the buffer here as well please

source/2_platform/rex_directx/src/directx_renderer.cpp
```diff
@@ -0,0 +1,137 @@
...
+
+    //-------------------------------------------------------------------------
+    bool is_depth_0_to_1()
+    {
+      return true;
+    }
+
+    namespace backend
+    {
+      struct DirectXContext
```
RisingLiberty, line 65: just letting you know you have a win_directx_context.cpp file in RexWindows, but the DirectXContext is defined here, not sure if that's what you wanted

RisingLiberty (changes requested):

source/2_platform/rex_directx/src/directx_renderer.cpp
Dyronix: Yes I know but this is actually a struct containing device dependent objects, I called it context cuz of lack of better naming

source/2_platform/rex_directx/src/directx_renderer.cpp
Dyronix: There is already an assert in the adaptor manager

source/2_platform/rex_directx/src/dxgi/adapter.cpp
```diff
@@ -0,0 +1,118 @@
...
+  const uint32 g_adapter_description_size = 128;
+
+  //-------------------------------------------------------------------------
+  /**
+   * The VendorId is a unique identifier assigned by the PCI-SIG (Peripheral Component Interconnect Special Interest Group)
+   * to identify the manufacturer of a PCI device, including graphics adapters.
+   * The VendorId is a 16-bit unsigned integer that is typically included in
+   * the PCI Configuration space of the device.
+   */
+  enum class Vendors
```
RisingLiberty, line 18: nit pick, but this is plural while it should be "Vendor" instead

source/2_platform/rex_directx/src/dxgi/adapter.cpp
RisingLiberty: this is still hasn't been done

RisingLiberty (changes requested):

RisingLiberty (dismissed):

RisingLiberty (approved):

RisingLiberty (approved):

PR #76, opened 2023-04-16
Development/assert system

RisingLiberty:


_test.py
```diff
@@ -17,6 +17,7 @@
 if __name__ == "__main__":
   parser = argparse.ArgumentParser()
   parser.add_argument("-clean", help="clean run, as if run for the first time", action="store_true")
+  parser.add_argument("-single_threaded", help="run tests in single threaded mode", action="store_true")
```
Dyronix, line 20: Is this only for the tests?
There is a define within rex called : REX_SINGLE_THREADED I was planning on configuring this macro via the build system. this will probably conflict with that. 
This is not something you should fix on this branch just keep it in mind, maybe we should create an issue for this.

source/1_engine/rex_engine/include/rex_engine/diagnostics/stacktrace.h
```diff
@@ -0,0 +1,5 @@
+#pragma once
+
+#ifdef REX_PLATFORM_WINDOWS
+  #include "rex_engine/diagnostics/win/win_stacktrace.h"
```
Dyronix, line 4: I think this file belongs in some platform layer no?
Since it's windows only, if this makes it to confusing you can leave it here just something to consider.
I posted a lot more comments like this within the review the same applies to them.

source/1_engine/rex_engine/include/rex_engine/diagnostics/win/win_stacktrace.h
```diff
@@ -0,0 +1,32 @@
+#pragma once
```
Dyronix, line 1: Platform layer?

source/1_engine/rex_engine/include/rex_engine/platform.h
```diff
@@ -0,0 +1,14 @@
+#pragma once
```
Dyronix, line 1: Not sure this is the correct location for this file unless it would be used for other purposes than the assert system I don't think this should be here.

source/1_engine/rex_engine/src/diagnostics/assert.cpp
```diff
@@ -0,0 +1,87 @@
...
+  DECLARE_LOG_CATEGORY_EXTERN(LogAssert, LogVerbosity::Log);
+  DEFINE_LOG_CATEGORY(LogAssert);
+
+  DebugVector<AssertContext>& contexts()
+  {
+    thread_local DebugVector<AssertContext> contexts;
+    return contexts;
+  }
+
+  void rex_assert(const rsl::fmt_stack_string& msg)
```
Dyronix, line 18: I'm going to need some explanation why this is such a complex thing.
Is there a reason why it does not log a message and breaks.

source/1_engine/rex_engine/src/diagnostics/win/win_stacktrace.cpp
```diff
@@ -0,0 +1,130 @@
+#include "rex_engine/diagnostics/win/win_stacktrace.h"
```
Dyronix, line 1: This file could use a lot more documentation, as this is very windows specific and not everybody will be able to read this easily if bugs occur.

Dyronix (changes requested):

source/1_engine/rex_engine/include/rex_engine/diagnostics/win/win_stacktrace.h
```diff
@@ -0,0 +1,32 @@
...
+#include "rex_std/array.h"
+#include "rex_std/format.h"
+
+#include "rex_std/stacktrace.h"
+
+namespace rex
+{
+  namespace win
+  {
+    inline constexpr card32 g_max_stack_entries = 20;
```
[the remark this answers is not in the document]
RisingLiberty: it's a constexpr variable, that does use inline.

source/1_engine/rex_engine/include/rex_engine/memory/memory_header.h
```diff
@@ -0,0 +1,32 @@
...
+
+    const win::CallStack& callstack() const;
+    rsl::memory_size size() const;
+    void* ptr() const;
+    rsl::thread::id thread_id() const;
+    MemoryTag tag() const;
+    card32 frame_index() const;
+
+  private:
+    win::CallStack m_callstack;
```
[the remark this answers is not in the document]
RisingLiberty: why? this is crossplatform

source/1_engine/rex_engine/include/rex_engine/memory/memory_stats.h
```diff
@@ -0,0 +1,37 @@
...
+
+    // amount of page faults that have occurred
+    // Note: this doesn't differentiate between hard and soft page faults
+    // so don't assume that a high page fault count is bad
+    card32 num_page_faults{};
+  };
+} // namespace rex
+
+#ifdef REX_PLATFORM_WINDOWS
+  #include "rex_engine/memory/win/win_mem_stats.h"
```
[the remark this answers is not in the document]
RisingLiberty: how so? this is the header you include in cross platform code, internally it checks the platform and the include the right header

source/1_engine/rex_engine/include/rex_engine/memory/win/win_mem_stats.h
```diff
@@ -0,0 +1,12 @@
+#pragma once
```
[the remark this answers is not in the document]
RisingLiberty: it is in a platform layer

_test.py
RisingLiberty: this is different, the single threaded is for the python scripts, it's not for rex

RisingLiberty: yes, it shouldn't be fixed in this branch, I thought it was fixed in async_ci

source/1_engine/rex_engine/src/diagnostics/win/win_stacktrace.cpp
RisingLiberty: it's in a "win" folder and has a "win" prefix, of course it's windows specific

source/1_engine/rex_engine/include/rex_engine/diagnostics/win/win_stacktrace.h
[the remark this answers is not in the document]
Dyronix: That is only one part of the issue

source/1_engine/rex_engine/include/rex_engine/memory/memory_stats.h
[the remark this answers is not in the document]
Dyronix: Then why is it within a PLATFORM macro?

source/1_engine/rex_engine/include/rex_engine/diagnostics/win/win_stacktrace.h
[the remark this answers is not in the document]
RisingLiberty: yes, making it configurable is fine, although I see little point in doing so as I don't think it'll change that often

_test.py
Dyronix: That's what I said that it's different, but it's going to conflict

RisingLiberty: then just rename the param in _test.py, I just need to know because teamcity uses it

source/1_engine/rex_engine/src/diagnostics/win/win_stacktrace.cpp
Dyronix: Yes so? 
I only asked to write documentation to the code.

Dyronix:
CI fails

Dyronix (changes requested):

source/1_engine/rex_engine/src/diagnostics/assert.cpp
```diff
@@ -0,0 +1,85 @@
...
+  {
+    thread_local static bool is_processing_assert = false;
```
Dyronix, line 20: Why does this need to be thread local?

source/1_engine/rex_engine/src/diagnostics/assert.cpp
```diff
@@ -0,0 +1,85 @@
...
+#include "rex_engine/diagnostics/stacktrace.h"
+
+namespace rex
+{
+  DECLARE_LOG_CATEGORY_EXTERN(LogAssert, LogVerbosity::Log);
+  DEFINE_LOG_CATEGORY(LogAssert);
+
+  DebugVector<AssertContext>& contexts()
+  {
+    thread_local DebugVector<AssertContext> contexts;
```
Dyronix, line 14: Why does this need to be thread local?

Dyronix (changes requested):

source/1_engine/rex_engine/include/rex_engine/diagnostics/assert.h
```diff
@@ -0,0 +1,52 @@
...
+  }
+#define REX_ASSERT_X(cond, ...)                                                                                                                                                                                                                          \
+  if(!(cond))                                                                                                                                                                                                                                            \
+  {                                                                                                                                                                                                                                                      \
+    REX_ASSERT(__VA_ARGS__);                                                                                                                                                                                                                             \
+  }
+
+#define REX_ASSERT_CONTEXT_SCOPE(...) rex::AssertContextScope ANONYMOUS_VARIABLE(assert_scope)(rsl::format(__VA_ARGS__));
```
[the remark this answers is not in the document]
RisingLiberty: it's not to replace asserts with, it's to provide extra information for asserts

source/1_engine/rex_engine/src/diagnostics/assert.cpp
RisingLiberty: because different threads could be processing an assert at the exact same time

source/1_engine/rex_engine/src/diagnostics/assert.cpp
RisingLiberty: because I didn't want to introduce locks into the system.
a thread owns what it's running and the context it's running in.
so when a thread asserts it doesn't care about the contexts on a different thread (at least not right now, this might change in the future)

Dyronix (approved):

PR #79, opened 2023-04-18
Development/stringid

Dyronix:


RisingLiberty:
please fix the issues with CI

source/1_engine/rex_engine/include/rex_engine/string/stringids.h
```diff
@@ -0,0 +1,20 @@
+#pragma once
+
+#include "rex_std/string.h"
+
+namespace rex
+{
+  enum class SID
```
RisingLiberty, line 7: not sure what this is for? is this ever needed?

source/1_engine/rex_engine/include/rex_engine/string/stringpool.h
```diff
@@ -0,0 +1,22 @@
...
+#include "rex_std/string.h"
+
+namespace rex
+{
+  class StringEntry;
+  class StringEntryID;
+
+  namespace string_pool
+  {
+    void resolve(const StringEntryID& entryID, const char** out, s32& outSize);
```
RisingLiberty, line 14: this can return a string_view

source/1_engine/rex_engine/include/rex_engine/string/stringpool.h
```diff
@@ -0,0 +1,22 @@
...
+    void resolve(const StringEntryID& entryID, rsl::string& out);
```
RisingLiberty, line 15: if the above returns a string_view, this can be removed

source/1_engine/rex_engine/include/rex_engine/string/stringpool.h
```diff
@@ -0,0 +1,22 @@
...
+
+    const StringEntry* find(const StringEntryID& entryID);
+
+    const StringEntryID* store(const SID& name);
+    const StringEntryID* store(const char* characters, s32 size);
```
RisingLiberty, line 20: use a string_view here

source/1_engine/rex_engine/include/rex_engine/string/stringentry.h
```diff
@@ -0,0 +1,65 @@
...
+#include "rex_engine/types.h"
+#include "rex_std/functional.h"
+#include "rex_std/memory.h"
+
+namespace rex
+{
+  class StringEntry
+  {
+  public:
+    StringEntry(const char* chars, s32 charCount) noexcept;
```
RisingLiberty, line 12: please take in a string_view instead

source/1_engine/rex_engine/src/string/stringid.cpp
```diff
@@ -0,0 +1,211 @@
...
+{
+  namespace internal
+  {
+    //-------------------------------------------------------------------------
+    StringEntryID make(const SID& name)
+    {
+      return name == SID::None ? StringEntryID() : *string_pool::store(name);
+    }
+    //-------------------------------------------------------------------------
+    StringEntryID make(const char* characters, s32 size)
```
RisingLiberty, line 17: this can take a string_view instead

source/1_engine/rex_engine/src/string/stringid.cpp
```diff
@@ -0,0 +1,211 @@
...
+//-------------------------------------------------------------------------
+rex::StringID operator""_sid(const char* string, size_t size)
+{
+  return rex::StringID(string, static_cast<u32>(size)); // NOLINT(cppcoreguidelines-narrowing-conversions)
+}
+
+//-------------------------------------------------------------------------
+rsl::ostream& operator<<(rsl::ostream& os, const rex::StringID& stringID)
+{
+  os << stringID.to_string();
```
RisingLiberty, line 208: one of the reasons why this should return a view.
otherwise you have a string allocation every time you call this which is exactly what you want to avoid with string pooling

source/1_engine/rex_engine/include/rex_engine/string/stringids.h
```diff
@@ -0,0 +1,20 @@
...
+    // Special zero value, meaning no name.
+    None,
+
+    // Special constant for the last hard-coded name index
+    MaxHardcodedNameIndex,
+  };
+
+  namespace conversions
+  {
+    rsl::string to_display_string(const SID& name);
```
RisingLiberty, line 18: this should return a string view

source/4_tools/regina/src/regina.cpp
```diff
@@ -31,7 +33,7 @@ namespace rex
 
     app_params.gui_params.window_width  = 1280;
     app_params.gui_params.window_height = 720;
-    app_params.gui_params.window_title  = "Regina";
+    app_params.gui_params.window_title  = "Regina"_sid;
```
RisingLiberty, line 36: doesn't really have to be a string id

source/1_engine/rex_engine/src/string/stringpool.cpp
```diff
@@ -0,0 +1,155 @@
...
+#include "rex_std/functional.h"
+#include "rex_std/internal/utility/pair.h"
+#include "rex_std/unordered_map.h"
+
+namespace rsl
+{
+  inline namespace v1
+  {
+    // https://en.wikipedia.org/wiki/Fowler–Noll–Vo_hash_function
+    rsl::hash_result hash_bytes(const void* data, rsl::count_t size) noexcept
```
RisingLiberty, line 14: we already have a hash implementation for bytes.
probably best if we stick with that so all string hashes are equivalenet to each other

source/1_engine/rex_engine/src/string/stringpool.cpp
```diff
@@ -0,0 +1,155 @@
...
+        hash ^= bytes[i];
+        hash *= fnv_prime;
+      }
+
+      return hash;
+    }
+
+    // Template specialization for hash a rsl::string
+    template <>
+    struct hash<rsl::string>
```
RisingLiberty, line 33: this already exists in rsl afaik, if not, then this should get added

source/1_engine/rex_engine/src/string/stringpool.cpp
```diff
@@ -0,0 +1,155 @@
...
+    {
+      rsl::hash_result operator()(const rsl::string& s) const noexcept
+      {
+        return hash_bytes(s.data(), s.size());
+      }
+    };
+
+    // Template specialization for hash a const char*
+    template <>
+    struct hash<const char*>
```
RisingLiberty, line 43: this already exists in rsl, if not, then this should get added

source/1_engine/rex_engine/src/string/stringpool.cpp
```diff
@@ -0,0 +1,155 @@
...
+    using EntryMap = rsl::unordered_map<StringEntryID, StringEntry>;
+
+    //-------------------------------------------------------------------------
+    EntryMap load_entry_map()
+    {
+      EntryMap map;
+
+      for(u32 i = 0; i < static_cast<u32>(SID::MaxHardcodedNameIndex); ++i)
+      {
+        rsl::string sid_name = conversions::to_display_string(static_cast<SID>(i));
```
RisingLiberty, line 66: should be string_view

source/1_engine/rex_engine/src/string/stringpool.cpp
```diff
@@ -0,0 +1,155 @@
...
+  namespace string_pool
+  {
+    using EntryMap = rsl::unordered_map<StringEntryID, StringEntry>;
+
+    //-------------------------------------------------------------------------
+    EntryMap load_entry_map()
+    {
+      EntryMap map;
+
+      for(u32 i = 0; i < static_cast<u32>(SID::MaxHardcodedNameIndex); ++i)
```
RisingLiberty, line 64: use enum reflection for this, but ideally I'm not sure the SID enum is necessary

source/1_engine/rex_engine/src/string/stringpool.cpp
```diff
@@ -0,0 +1,155 @@
...
+      if(it == rsl::cend(get_entries()))
+      {
+        const rsl::hash_result sid_name = static_cast<rsl::hash_result>(SID::None);
+
+        it = get_entries().find(sid_name);
+
+        REX_ASSERT_X(it != rsl::cend(get_entries()), "SID::None not present");
+      }
+
+      return &(it->value);
```
RisingLiberty, line 139: use rsl::iterator_to_pointer here

source/1_engine/rex_engine/src/string/stringpool.cpp
```diff
@@ -0,0 +1,155 @@
...
+
+    //-------------------------------------------------------------------------
+    EntryMap& get_entries()
+    {
+      static EntryMap entries = load_entry_map();
+      return entries;
+    }
+
+    //-------------------------------------------------------------------------
+    const StringEntryID* store(rsl::hash_result hash, const char* characters, s32 size)
```
RisingLiberty, line 85: as specified in the header, more string_view usage is needed here

RisingLiberty (changes requested):

source/1_engine/rex_engine/include/rex_engine/string/stringids.h
Dyronix: Yes to initialize a StringID to 0 but I can make it a state value does not have to be an enum

RisingLiberty: why is a default construced StringID not initialized with 0?

RisingLiberty: well I'd like to check it with neither really.
why don't you have a "empty()" func just like a regular string?

Dyronix: For the same reason I like to check my booleans with == false and == true

Dyronix: It's more readably IMO

Dyronix: I can make it some constant does not have to be an enum, this was only there if there would be any other hardcoded values

RisingLiberty: what is it actually that you want to test? if the StringID exists?
StringIDs should have the same interface as actual strings, and all string like interfaces already have "empty" func in there to check if there's actually something in it, now you can have a StringID that is empty so I suggest to use an interface func like "is_valid" or something.

RisingLiberty: hardcoded values shouldn't be StringIDs, why can't they be string views?

Dyronix: A StringID will never be empty, as it would store 0 or some other value, 0 is also a valid input so is_valid will also not really work. I'm not sure why StringID should have the same API as a string it's not a string. it's a hashed value to a string. For instance, you cannot query the size without looking up the actual string, you cannot iterate over it as with a string. That's why I would like to explicitly check if the value within the StringID == None as with empty or is_valid you don't really know what value is in it. You already have to know the API internally to know that empty or is_valid means None. StringID's do not store string_views they store ints and to compare a StringID with a string_view a lookup has to be made within the stringpool in order to check if it's similar which is where you will compare 2 strings instead of 2 integers if you compare with another StringID ( hardcoded or not )

Dyronix: I already have an "is_none" function though, this checks on an empty SID
  /** True for StringID(), StringID(NAME_None) and StringID("None") */
  bool StringID::is_none() const
  {
    return *this == SID::None;
  }

source/1_engine/rex_engine/src/string/stringpool.cpp
Dyronix: Could you tell me where I can find it?

source/1_engine/rex_engine/src/string/stringpool.cpp
Dyronix: Could you tell me where I can find it?

source/1_engine/rex_engine/src/string/stringpool.cpp
Dyronix: Why would I add another indirection here?

      return &rsl::iterator_to_pointer(it)->value;
vs
      return &(it->value);

source/1_engine/rex_engine/include/rex_engine/string/stringid.h
```diff
@@ -0,0 +1,60 @@
...
+
+  bool operator==(rsl::string_view s, const StringID& sid);
+  bool operator!=(rsl::string_view s, const StringID& sid);
+  bool operator==(const StringID& sid, rsl::string_view s);
+  bool operator!=(const StringID& sid, rsl::string_view s);
+} // namespace rex
+
+rex::StringID operator""_sid(const char* string, size_t size);
+
+rsl::ostream& operator<<(rsl::ostream& os, const rex::StringID& stringID);
```
RisingLiberty, line 43: probably best if you create a formatter for this as well.
I'll see if I can create a default formatter to work with rsl::format that uses the operator<< overload for ostream

source/1_engine/rex_engine/src/string/stringentry.cpp
```diff
@@ -0,0 +1,97 @@
...
+
+#include <rex_std/memory.h>
+
+namespace rex
+{
+  //-------------------------------------------------------------------------
+  StringEntry::StringEntry(rsl::string_view characters) noexcept
+      : m_characters()
+  {
+    m_characters.reset(static_cast<char*>(malloc(characters.size() + 1))); // NOLINT(cppcoreguidelines-no-malloc)
```
RisingLiberty, line 11: don't use malloc, the memory won't get tracked if you do.
just a note, you can use rsl::make_unique<char[]>(characters.size() + 1); here
that'll use operator new internally and will track memory.
cppcoreguidelines-no-malloc should never be ignored unless approved by both you and me

source/1_engine/rex_engine/src/string/stringentry.cpp
```diff
@@ -0,0 +1,97 @@
...
+
+    rsl::memset(m_characters.get(), 0, characters.size() + 1);
```
RisingLiberty, line 13: why do you set to 0 first, only to copy in it afterwards?

source/1_engine/rex_engine/src/string/stringentry.cpp
```diff
@@ -0,0 +1,97 @@
...
+  {
+    m_characters = rsl::exchange(other.m_characters, {});
+
+    return *this;
+  }
+
+  //-------------------------------------------------------------------------
+  rsl::string_view StringEntry::characters() const
+  {
+    return rsl::string_view(m_characters.get());
```
RisingLiberty, line 36: provide the length as well, this saves a strlen call in the ctor
general rule is, if you have the length, pass it in

source/1_engine/rex_engine/src/string/stringid.cpp
```diff
@@ -0,0 +1,137 @@
...
+#include "rex_std/vector.h"
+
+namespace rex
+{
+  namespace internal
+  {
+    //-------------------------------------------------------------------------
+    StringEntryID make(rsl::string_view characters)
+    {
+      if(characters.empty())
```
RisingLiberty, line 14: should we warn about this?

source/1_engine/rex_engine/src/string/stringid.cpp
```diff
@@ -0,0 +1,137 @@
...
+      return *string_pool::store(characters);
+    }
+  } // namespace internal
+
+  //-------------------------------------------------------------------------
+  /**
+   * Create an empty StringID.
+   */
+  StringID::StringID()
+      : m_comparison_index(0)
```
RisingLiberty, line 28: can this use an internal invalid id?
it's hashing so it's never be 0 afaik but a proper state that this is an invalid id is probably useful here

source/1_engine/rex_engine/include/rex_engine/string/stringid.h
```diff
@@ -0,0 +1,60 @@
...
+    bool operator==(const StringEntryID& entryID) const;
+    bool operator!=(const StringEntryID& entryID) const;
+
+    rsl::string_view to_string_view() const;
+    bool is_none() const;
+    u32 value() const;
+
+  private:
+    /** Index into the StringID array (used to find String portion of the string/number pair used for comparison) */
+    StringEntryID m_comparison_index;
```
RisingLiberty, line 30: this is not really an index but a key into the string pool which is a hash map.
because you used the term "index" I thought it was an array

source/1_engine/rex_engine/src/string/stringid.cpp
```diff
@@ -0,0 +1,137 @@
...
+  rsl::string_view StringID::to_string_view() const
+  {
+    return string_pool::resolve(m_comparison_index);
+  }
+
+  //-------------------------------------------------------------------------
+  /** True for StringID(), StringID(NAME_None) and StringID("None") */
+  bool StringID::is_none() const
+  {
+    return m_comparison_index == 0;
```
RisingLiberty, line 85: can this use an internal invalid id?
it's hashing so it's never be 0 afaik but a proper state that this is an invalid id is probably useful here
I know we're not using SID::None anymore, which I still believe is fine as that shouldn't be exposed to the client.
but internally, I do think it's worth keeping an invalid hash result value making these statements easier to spot when looking at the internal code

source/1_engine/rex_engine/src/string/stringpool.cpp
```diff
@@ -0,0 +1,88 @@
...
+      StringEntryID entry_id(hash);
+      StringEntry entry(characters);
+
+      auto result = get_entries().emplace(rsl::move(entry_id), rsl::move(entry));
+      if(result.emplace_successful)
+      {
+        return &result.inserted_element->key;
+      }
+
+      return nullptr;
```
RisingLiberty, line 55: don't you want to error here?
alternatively, in theory I can't think of a way (other than running out of memory) where this branch would get executed.
a hash collision is caught above, the only other reason why it could fail I believe is when it fails to allocate a new entry into the hash map (but even then, it'd return a successful emplace result, it'll likely crash before reaching that point though)

so is it worth returning a pointer and no the element directly?

RisingLiberty (changes requested):

source/1_engine/rex_engine/src/string/stringpool.cpp
Dyronix: cuz store also returns a ptr and this one can fail and I wanted to have a consistent API

source/1_engine/rex_engine/include/rex_engine/string/stringid.h
```diff
@@ -0,0 +1,60 @@
...
+  };
+
+  StringID create_sid(rsl::string_view stringView);
+
+  bool operator==(rsl::string_view s, const StringID& sid);
```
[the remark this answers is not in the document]
Dyronix: I have somewhat of a coding convention that everything which is a class or struct I pass by ref, I can pass it by value if you want. The reason why this is is because if the class/struct changes I do not have to worry I am making expensive copies.

source/1_engine/rex_engine/include/rex_engine/string/stringid.h
```diff
@@ -0,0 +1,60 @@
...
+    rsl::string_view to_string_view() const;
+    bool is_none() const;
+    u32 value() const;
+
+  private:
+    /** Index into the StringID array (used to find String portion of the string/number pair used for comparison) */
+    StringEntryID m_comparison_index;
+  };
+
+  StringID create_sid(rsl::string_view stringView);
```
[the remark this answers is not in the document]
Dyronix: there is no way to remove string_ids that is true but I don't think this is required though. what might be useful is a clear or something. for instance, if we switch between different states of the application it might be an idea to free up some memory for strings that will not be used. however, we do need to keep in mind if we switch back to the other state the string pool has to be filled again (unless we store it somewhere on disk)

source/1_engine/rex_engine/src/string/stringentry.cpp
Dyronix: Okay, thanks!

source/1_engine/rex_engine/src/string/stringpool.cpp
RisingLiberty: I also put a comment in store if that should return a pointer

source/1_engine/rex_engine/include/rex_engine/string/stringid.h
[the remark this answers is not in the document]
RisingLiberty: there's a class in rsl that makes it a reference if it's bigger than 1 pointer size, a non-ref if not if you're interested

source/1_engine/rex_engine/src/string/stringentry.cpp
Dyronix: Cuz it's zero terminated, IIRC it's faster to clear everything to 0 and copy the data you want over first copy the string and place the 0 afterwards, also I think this looks cleaner.

source/1_engine/rex_engine/include/rex_engine/string/stringid.h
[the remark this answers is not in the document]
RisingLiberty: I think it's worth to keep in mind but doesn't have to be implemented in this change

source/1_engine/rex_engine/src/string/stringid.cpp
Dyronix: Not sure, the empty one is capture with the default ctor from StringID, you only enter the ::make function when you try and store a string. so perhaps I leave this up to you.

source/1_engine/rex_engine/src/string/stringentry.cpp
RisingLiberty: can you put a comment why you do that then?
I'm not sure if it's faster, it'll probably depend on the length of the string.
it'll also depend if this first access to the memory, on first access and if the memory used for this was on a previous unused page, the OS needs to actually initialize the page, that might be what you've seen before.

I'd like to see this profiled in possible though, if you copy everything first and then add a null termination char at the end, I dont' see how this would be slower over what's implemented here

source/1_engine/rex_engine/src/string/stringid.cpp
Dyronix: It's not an invalid state, it's the none state. meaning this is empty.

source/1_engine/rex_engine/src/string/stringid.cpp
Dyronix: So what do you propose?

source/1_engine/rex_engine/include/rex_engine/string/stringid.h
Dyronix: So you want me to rename it or is this just a general comment?

source/1_engine/rex_engine/src/string/stringid.cpp
RisingLiberty: the default one doesn't go into make so I'd say we should error or at least warn here as this means a StringID is used on an empty string (StringID are used on data, not sure why we need an empty one there). If we discover a case where a empty string used by a string id makes sense, we can revisit this

RisingLiberty: better to put constrains in first and release them later than the other way around

source/1_engine/rex_engine/src/string/stringpool.cpp
Dyronix: It's there for safety reasons, I could **not** check if the element was inserted successfully but that would be a nasty bug if it would ever fail .. Not sure when but better safe than sorry.

source/1_engine/rex_engine/src/string/stringid.cpp
RisingLiberty: what exactly is an empty string id definition?
a string id is an id that identifies a string, right?
in this case, it doesn't identify any string, which is a special case.
if that's a valid state or not doesn't really matter, I still think a proper named value to assign here is worth it

source/1_engine/rex_engine/include/rex_engine/string/stringid.h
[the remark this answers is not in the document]
Dyronix: Which is?

source/1_engine/rex_engine/src/string/stringid.cpp
RisingLiberty: I propose that StringID has an internal constexpr static member s_none_state_hash = 0 (or something similarly named)
and all StringIDs initialized with this get initialized with this value.
this also makes it possible to easier debug string ids in the future

source/1_engine/rex_engine/include/rex_engine/string/stringid.h
Dyronix: Made an issue for it, so I can figure out how to do this but the branch can already be merged. It's an addition to the API not a change so I think this it's fine to do it after.

source/1_engine/rex_engine/src/string/stringpool.cpp
RisingLiberty: would an assert above the return that emplacement failed and returning a default constructed StringEntryID satisfy you in this case or do you prefer the nullptr?
PS: I think it's worth adding an assert regardless

source/1_engine/rex_engine/include/rex_engine/string/stringid.h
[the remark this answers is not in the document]
RisingLiberty: rsl::call_traits<T>::param_type
a bit verbose, I know

source/1_engine/rex_engine/src/string/stringid.cpp
Dyronix: The 0 is a hash to "Invalid String" so if you would call toString on an empty ( aka 0 is stored inside the comparison index ) you retrieve "Invalid String" so in theory it's not empty. I can store some special value if you want but is 0 not good enough?

source/1_engine/rex_engine/src/string/stringpool.cpp
Dyronix: It would satisfy me yes. Although I wonder why this pointer is bothering you?

source/1_engine/rex_engine/src/string/stringid.cpp
RisingLiberty: 0 is fine, I just want it behind a name

source/1_engine/rex_engine/include/rex_engine/string/stringid.h
RisingLiberty: please rename it

RisingLiberty: also update the comment above it "Index into the StringID array"
it's not an array, it's hash table

source/1_engine/rex_engine/src/string/stringpool.cpp
RisingLiberty: it's an indirection

source/1_engine/rex_engine/src/string/stringentry.cpp
RisingLiberty: setting only the last char is always faster
https://github.com/New-Creta/Rex/issues/103

RisingLiberty (changes requested):

source/1_engine/rex_engine/src/string/stringid.cpp
Dyronix: Yes it is.
image

source/1_engine/rex_engine/src/string/stringpool.cpp
Dyronix: Issue has been created for this, it's not causing any issues now.
https://github.com/New-Creta/Rex/issues/105

source/1_engine/rex_engine/src/string/stringpool.cpp
```diff
@@ -0,0 +1,89 @@
...
+  namespace string_pool
+  {
+    using EntryMap = rsl::unordered_map<StringEntryID, StringEntry>;
+
+    //-------------------------------------------------------------------------
+    EntryMap load_entry_map()
+    {
+      EntryMap map;
+
+      StringEntryID entry_id(0);
```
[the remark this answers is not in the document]
Dyronix: Well it's not a StringID, it's the StringEntryID not the same class, so not sure if this should have been applied here as well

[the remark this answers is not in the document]
RisingLiberty: I'd say so, if for some reason that hardcoded value would change, the above could would be broken, there's a dependency here that s_none_state_hash == 0 here, so might as well use the variable

RisingLiberty (approved):

PR #87, opened 2023-04-29
Development/event system

Dyronix:


source/1_engine/rex_engine/include/rex_engine/core_application.h
```diff
@@ -60,6 +63,7 @@ namespace rex
     void loop();
 
   private:
+    bool m_is_paused;
```
RisingLiberty, line 66: I think it's better to have an enum for the state of the application instead of adding flags.
in the future we might have more complex states

source/1_engine/rex_engine/include/rex_engine/event.h
```diff
@@ -0,0 +1,22 @@
...
+{
+  namespace event_system
+  {
+    struct WindowReizeEvent
+    {
+      s32 window_width;
+      s32 window_height;
+    };
+
+    union Event
```
RisingLiberty, line 16: why is this a union? it looks like both these fields will be used at the same time which will be UB at that point.

source/1_engine/rex_engine/src/event.cpp
```diff
@@ -0,0 +1,9 @@
+#include "rex_engine/event.h"
```
RisingLiberty, line 1: please don't add empty cpp files

source/1_engine/rex_engine/src/event_type.cpp
```diff
@@ -0,0 +1,9 @@
+#include "rex_engine/event_type.h"
```
RisingLiberty, line 1: please don't add empty cpp files

source/2_platform/rex_windows/src/gui_application.cpp
```diff
@@ -15,12 +15,46 @@
...
 
-// NOLINTBEGIN(modernize-use-nullptr,-warnings-as-errors)
+// NOLINTBEGIN(cppcoreguidelines-pro-type-union-access)
+// NOLINTBEGIN(modernize-use-nullptr)
 
 namespace rex
 {
   namespace win32
   {
+    class GuiApplicationState
```
RisingLiberty, line 25: this is a window state, not an application state

source/2_platform/rex_windows/src/gui_application.cpp
```diff
@@ -15,12 +15,46 @@
...
+        Minimized = BIT(1),
+        Maximized = BIT(2)
+      };
+
+      GuiApplicationState()
+          : m_state(0u)
+      {
+      }
+
+      void add_state(GuiApplicationState::Type state)
```
RisingLiberty, line 40: these functions look like they'd be more benefitial as library functions so they can be used for any enum

source/2_platform/rex_windows/src/gui_application.cpp
```diff
@@ -15,12 +15,46 @@
...
+      {
+        m_state = m_state & static_cast<u32>(state);
+      }
+      void remove_state(GuiApplicationState::Type state)
+      {
+        m_state = m_state & ~static_cast<u32>(state);
+      }
+
+      bool has_state(GuiApplicationState::Type state) const
```
RisingLiberty, line 49: this already exists. rsl::has_flag

source/2_platform/rex_windows/src/gui_application.cpp
```diff
@@ -142,6 +187,101 @@ namespace rex
         }
       }
 
+      void start_resize()
```
RisingLiberty, line 190: maybe best to specify the "on_" prefix for all callbacks for events.
makes them easier to distinquish from other functions

source/2_platform/rex_windows/src/gui_application.cpp
```diff
@@ -142,6 +187,101 @@ namespace rex
...
+        // Resize window ( although we might want to capture this within the window itself ... )
+        //
+        // Flush Command Queue
+        //
+        // Release front and back buffer
+        // Release depth stencil buffer
+        //
+        // Reset current back buffer back to first entry ( aka 0 )
+        //
+        // Recreate swap chain buffers
```
RisingLiberty, line 272: I think you only want to do this on end resize not during resizing.
you want to stop rendering on resize start and resume rendering on resize finish.
no rendering needs to happen during resizing

RisingLiberty (changes requested):

source/1_engine/rex_engine/include/rex_engine/event.h
Dyronix: I have not tested it on clang but this is how SDL handles it's events and it seems quite alright here.

source/2_platform/rex_windows/src/gui_application.cpp
Dyronix: This is not present in rsl I checked.
RSL expects the first argument to be an integer and the second to be an enum which is the other way around here.

PR #111, opened 2023-05-09
[bug/adapter_desc] fixing bugs in adapter desc retrieval

RisingLiberty:


Dyronix (approved):
This is not really a bug but more a boyscout ...

PR #113, opened 2023-05-13
[development/console_application] added console application API on engine side

RisingLiberty:


source/2_platform/rex_windows/src/console_application.cpp
```diff
@@ -0,0 +1,114 @@
...
+
+#include <Windows.h>
+
+namespace rex
+{
+  namespace win32
+  {
+    DEFINE_LOG_CATEGORY(LogWinConsoleApp, rex::LogVerbosity::Log);
+
+    CoreApplication* g_this_app = nullptr; // NOLINT(cppcoreguidelines-avoid-non-const-global-variables)
```
Dyronix, line 14: What is the difference between this ptr and the app_instance ptr?
Can't we just use one ptr?

source/1_engine/rex_engine/src/entrypoint.cpp
```diff
@@ -0,0 +1,10 @@
+#include "rex_engine/entrypoint.h"
+
+#include "rex_std/string.h"
+
+namespace rex
+{
+  void pre_app_entry(REX_MAYBE_UNUSED const char8* cmdLine) {}
+
+  void post_app_shutdown() {}
```
Dyronix, line 9: I'm not a big fan of the pre and post design ... this always has to be executed in order and is easy to mess up IMO.

source/2_platform/rex_windows/include/rex_windows/platform_creation_params.h
```diff
@@ -14,4 +15,34 @@ namespace rex
...
   };
+
+  struct GuiParams
+  {
+    s32 window_width      = 1280;
+    s32 window_height     = 720;
+    StringID window_title = "Application"_sid;
+
+    s32 max_render_commands = 1 << 16;
+    s32 max_fps             = 60;
```
Dyronix, line 26: Is there an issue for these that we should move them into a Config/engine_settings.ini?

source/2_platform/rex_windows/src/console_application.cpp
```diff
@@ -0,0 +1,114 @@
...
+        m_on_initialize = m_engine_params.app_init_func ? m_engine_params.app_init_func : [&]() { return true; };
+
+        m_on_update = m_engine_params.app_update_func ? m_engine_params.app_update_func : [&]() {};
+
+        m_on_shutdown = m_engine_params.app_shutdown_func ? m_engine_params.app_shutdown_func : [&]() {};
+      }
+
+      bool initialize()
+      {
+        SetConsoleCtrlHandler(handler_routine, true); // NOLINT(readability-implicit-bool-conversion)
```
Dyronix, line 54: Would it be possible to create a constexpr bool here and pass in that to the SetConsoleCtrlHandler, if you give the variable a good name we know what the "true" stands for.

source/2_platform/rex_windows/src/console_application.cpp
```diff
@@ -0,0 +1,114 @@
...
+
+        event_system::subscribe(event_system::EventType::QuitApp, [this](const event_system::Event& /*event*/) { m_app_instance->quit(); });
```
Dyronix, line 56: I know you explained me why we need a QuitApp event but could you add a comment here ( or with the event system )  why we need this. As we do have a WindowClose but I remembered this wasn't good enough in this case just cannot recall why.

source/2_platform/rex_windows/src/console_application.cpp
```diff
@@ -0,0 +1,114 @@
...
+
+    ConsoleApplication::ConsoleApplication(ApplicationCreationParams&& appParams)
+        : CoreApplication(appParams.engine_params, appParams.cmd_args)
+        , m_internal_obj()
+        , m_internal(m_internal_obj.get<ConsoleApplication::Internal>())
+    {
+      static_assert(s_internal_size == sizeof(ConsoleApplication::Internal), "size of internal structure should match size of internal implementation");
+      static_assert(s_internal_alignment == alignof(ConsoleApplication::Internal), "alignment of internal structure should match alignment of internal implementation");
+
+      new(m_internal)(ConsoleApplication::Internal)(this, rsl::move(appParams));
```
Dyronix, line 95: This really seems like overkill to me ... 
Why does this need to be this complex?
We only have 1 console app so the one allocation is not an issue here.
Readability should always triumph over verbosity only when this poses a real issue we should consider optimizing this.

source/1_engine/rex_engine/src/entrypoint.cpp
Dyronix: Is post_app_shutdown even used?

Dyronix (changes requested):

source/1_engine/rex_engine/src/entrypoint.cpp
RisingLiberty: How is it easy to mess up? It's literally in the name when you have to use it?
And it's the engine that calls it so the client shouldn't worry about it

source/2_platform/rex_windows/include/rex_windows/console_application.h
```diff
@@ -0,0 +1,35 @@
...
+      bool platform_init() override;
+      void platform_update() override;
+      void platform_shutdown() override;
+
+    private:
+      class Internal;
+      static constexpr card32 s_internal_size      = 264;
+      static constexpr card32 s_internal_alignment = 8;
+      rsl::aligned_storage<s_internal_size, s_internal_alignment> m_internal_obj;
+      Internal* m_internal;
```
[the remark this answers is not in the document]
RisingLiberty: It's for easy of use

source/2_platform/rex_windows/include/rex_windows/platform_creation_params.h
RisingLiberty: There's an issue for engine config

source/2_platform/rex_windows/src/console_application.cpp
RisingLiberty: The reason why I made it stack and not heap memory is to avoid cluttering the memory tracking these kind of allocations.
Another way would be to use the untracked allocator

source/2_platform/rex_windows/src/console_application.cpp
RisingLiberty: there is no app instance anymore.

Dyronix: Why not?

source/1_engine/rex_engine/src/entrypoint.cpp
Dyronix: This is the console app ...

RisingLiberty: they're used with the file system

Dyronix: Anything that requires an order in which something is called can be messed up. It feels iffy =/ don't add them in global space encapsulate them and have one function queue them

source/2_platform/rex_windows/include/rex_windows/console_application.h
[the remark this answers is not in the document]
Dyronix: What's wrong with rsl::unique_ptr

source/1_engine/rex_engine/src/entrypoint.cpp
RisingLiberty: it's well defined as its marked with "pre" and "post" so I don't agree with that it's easy to mess up.

source/2_platform/rex_windows/include/rex_windows/console_application.h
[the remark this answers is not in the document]
RisingLiberty: It's not heap allocated

source/2_platform/rex_windows/src/console_application.cpp
Dyronix: Can't we add a tag for pimpl or internal? We have a memory tagging system why not use it?

RisingLiberty: Yes we can use a tag for that

source/1_engine/rex_engine/src/entrypoint.cpp
Dyronix: We had this structure in the past within the application with pre init and post init, I don't agree with how this is done. Encapsulate them within a class or a function an let the class call/function them in the right order. Just as I did for the context

source/2_platform/rex_windows/include/rex_windows/console_application.h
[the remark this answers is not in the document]
Dyronix: I think you meant that unique_ptr is a heap allocation, but that doesn't matter in this case IMO, this is only one allocation for a lot more readable code.

source/1_engine/rex_engine/src/entrypoint.cpp
Dyronix: Why do you need them, what do we require before we enter the application?

Dyronix: Or after we enter the application?

RisingLiberty: vfs initialization.
I've told you this before

Dyronix: And why can't we initialize this in CoreApplication?

RisingLiberty: because by then it's too late.
the client might want to access files for its app creation params.
they can't do that if vfs is not initialized by that point yet

Dyronix: But initialization of the VFS is our concern, the client has nothing to do with this right? So can't we make a function within entrypoint.cpp (and thus not expose it in the header ) that would initialize the VFS?

Dyronix: Then a client can still use the VFS in app_entry

source/2_platform/rex_windows/src/console_application.cpp
RisingLiberty: because it wasn't needed at the time, so I removed it.
if we want a generic design back, that's fine, but that's out of scope of this change imo

source/2_platform/rex_windows/src/console_application.cpp
RisingLiberty: I think I added a comment in another branch so to avoid merge conflicts, I won't do it here, just comment on this PR.
the reason why we have a QuitApp event is because WindowClose event != QuitApp event.
usually one does result in the other, but not always. and because a console application cannot hook into window close events of the console (at least not directly) 
this is the reason why we split them up into 2 different events.

source/2_platform/rex_windows/src/console_application.cpp
Dyronix: So to make things clear we had a feature that worked in the long run and you removed that and replaced it with something that works short term, isn't that against everything you are telling me?

RisingLiberty: we discussed together if we should remove it or not.
I believe I asked you if you were okay with removing it as we didn't see a point in having it (it's too exposed as it can be accessed everywhere) so we deleted it

Dyronix (approved):

PR #114, opened 2023-05-13
[devops/mem_tracking_improvements] made some improvements to mem tracking

RisingLiberty:


source/1_engine/rex_engine/rex_engine.sharpmake.cs
```diff
@@ -25,9 +26,11 @@ public RexEngine() : base()
     string relative_source_path = Util.PathGetRelative(Path.Combine(Globals.SourceRoot), SourceRootPath);
     GenerationConfigPath = Path.Combine(Globals.Root, "config", relative_source_path, "generation.json");
     MemoryTagsHeaderFile = Path.Combine(SourceRootPath, "include", "rex_engine", "memory", "memory_tags.h");
+    FileSystemMountHeaderFile = Path.Combine(SourceRootPath, "include", "rex_engine", "filesystem", "mount_roots.h");
```
Dyronix, line 29: mounting_points.h?

source/1_engine/rex_engine/rex_engine.sharpmake.cs
```diff
@@ -25,9 +26,11 @@ public RexEngine() : base()
...
 
     // create it here so it's added to the project in case it didn't exist yet
-    CreateMemoryTagFile();
+    TouchGenerationFile(MemoryTagsHeaderFile);
+    TouchGenerationFile(FileSystemMountHeaderFile);
```
Dyronix, line 33: I wonder how many times this would be used as I can only think of 2 things that would be mounted. Content and Config files. Are there different paths you would like to mount?

source/1_engine/rex_engine/src/memory/memory_tracking.cpp
```diff
@@ -17,27 +23,79 @@ namespace rex
...
   {
-    thread_local static rsl::array<MemoryTag, g_max_allowed_mem_tags> stack = {MemoryTag::Global};
+    thread_local static rsl::array<MemoryTag, g_max_allowed_mem_tags> stack = { MemoryTag::Global };
     return stack;
   }
 
+  class AllocationCallStack
+  {
+  public:
+    AllocationCallStack(CallStack callstack, card64 size)
```
Dyronix, line 33: We want to copy by value here, shouldn't we at least move them into the datamember then?

source/1_engine/rex_engine/src/memory/memory_tracking.cpp
```diff
@@ -17,27 +23,79 @@ namespace rex
...
+    return alloc_info_table;
+  }
+
   card32& thread_local_mem_tag_index()
   {
     thread_local static card32 tag = 0;
     return tag;
   }
 
   MemoryTracker::MemoryTracker()
```
Dyronix, line 96: Can we maybe dump the stats always in a debug build? cuz now you made this entire system but it does not have much usage. If you made it you probably have ideas about what you want it to do could you please also apply those? The only feature I see being used is the "initialize".

Dyronix (changes requested):

source/1_engine/rex_engine/include/rex_engine/memory/memory_tracking.h
```diff
@@ -32,17 +38,23 @@ namespace rex
 
     void initialize(rsl::memory_size maxMemUsage);
 
-    void track_alloc(void* mem, MemoryHeader* header);
-    void track_dealloc(void* mem, MemoryHeader* header);
+    MemoryHeader* track_alloc(void* mem, card64 size);
```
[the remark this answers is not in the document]
RisingLiberty: no we can't, blob is owner of its memory, that's no the case here.
blob is used for a blob of data, that it owns, that the client can do something with, either reading or writing.

source/1_engine/rex_engine/rex_engine.sharpmake.cs
RisingLiberty: yes, but more on that in the file system PR

source/1_engine/rex_engine/src/memory/memory_tracking.cpp
RisingLiberty: I can move it, making it future proof, but moving has no effect here

source/1_engine/rex_engine/src/memory/memory_tracking.cpp
RisingLiberty: the memory stats should only be dumped on demand, they shouldn't get dumped every frame or anything like that.
they will be part of an autotest and also for debugging purposes when you want to track memory and where it's going.
another idea is to create a tracking tool which can track memory over a longer time period (days, weeks, ... depending on our storage capabilities)

source/1_engine/rex_engine/include/rex_engine/memory/memory_tracking.h
```diff
@@ -32,17 +38,23 @@ namespace rex
...
+    void track_dealloc(MemoryHeader* header);
 
     void push_tag(MemoryTag tag);
     void pop_tag();
 
     MemoryTag current_tag() const;
 
-    MemoryUsageStats current_stats(); // deliberate copy as we don't want to have any race conditions when accessing
+    void dump_stats_to_file(rsl::string_view filepath);
```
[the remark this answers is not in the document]
Dyronix: Made an issue for this
https://github.com/New-Creta/Rex/issues/135

source/1_engine/rex_engine/src/memory/memory_tracking.cpp
Dyronix: Why do you think they should only be dumped on demand, if I had a difficult run which is hard to reproduce I can at least look at the memory dump that was automatically generated without me having to think about enabling it. Going to mark this as resolved but we should discuss this.

Dyronix (approved):

PR #118, opened 2023-05-13
[development/filesystem] added a file system API

RisingLiberty:


build/sharpmake/src/base.cs
```diff
@@ -254,6 +254,11 @@ public override void Configure(RexConfiguration conf, RexTarget target)
     }
   }
 
+  protected string GetClangToolsPath(RexConfiguration conf)
+  {
+    return Path.Combine(conf.ProjectPath, "clang_tools", conf.Target.GetFragment<Compiler>().ToString(), conf.Name);
```
Dyronix, line 259: Why does the file system API require the ClangToolsPath?

build/sharpmake/src/base.cs
```diff
@@ -309,11 +314,38 @@ private void ReadGenerationConfigFile()
...
+        }
+
+        if (enum_gen_settings.Filepath != enum_config.Filepath)
+        {
+          throw new Error($"Enum generation error - unexpected filepath: '{enum_config.Filepath}' - expected: {enum_gen_settings.Filepath} for project: {Name}");
+        }
+      }
+
+      EnumGenerationSettings enum_gen_setting = GenerateSettings.EnumsToAutoGenerate[key];
+      enum_gen_setting.ProjectToEnumValues.Add(Name, enum_config.Values);
```
Dyronix, line 348: Why does the file system API require the Memory Tags to be updated, please move boy scouting changes into their own branch so we don't confuse it with other features.

source/1_engine/rex_engine/include/rex_engine/filesystem/vfs.h
```diff
@@ -0,0 +1,70 @@
...
+    ReadRequest(rsl::string_view filepath, QueuedRequest* queuedRequest);
+    ReadRequest(const ReadRequest& other);
+    ReadRequest(ReadRequest&& other);
+
+    ~ReadRequest();
+
+    ReadRequest& operator=(const ReadRequest& other);
+    ReadRequest& operator=(ReadRequest&& other);
+
+    void signal(const char8* buffer, count_t count);
```
Dyronix, line 31: Use memory::blob

source/1_engine/rex_engine/include/rex_engine/filesystem/vfs.h
```diff
@@ -0,0 +1,70 @@
...
+    void wait() const;
+
+    const char8* data() const;
+    count_t count() const;
```
Dyronix, line 35: Use memory::blob

source/1_engine/rex_engine/include/rex_engine/filesystem/vfs.h
```diff
@@ -0,0 +1,70 @@
...
+
+    rsl::string_view filepath() const;
+
+  private:
+    rsl::string_view m_filepath;
+    QueuedRequest* m_queued_request;
+    bool m_is_done;
+    const char8* m_buffer;
+    count_t m_count;
```
Dyronix, line 44: Use memory::blob

source/1_engine/rex_engine/include/rex_engine/filesystem/vfs.h
```diff
@@ -0,0 +1,70 @@
...
+#include "rex_engine/types.h"
+#include "rex_std/bonus/functional.h"
+#include "rex_std/functional.h"
+#include "rex_std/memory.h"
+#include "rex_std/string_view.h"
+#include "rex_std/type_traits.h"
+#include "rex_std_extra/utility/type_id.h"
+#include "rex_std_extra/utility/yes_no.h"
+
+namespace rex::vfs
```
Dyronix, line 13: I don't like this style of namespaces, could we add nested ones?
I think we already do nested ones in different places.
This adds to readability as well cuz I can see the indentation an I know that this code is included within more namespaces than just "rex"

source/1_engine/rex_engine/include/rex_engine/filesystem/vfs.h
```diff
@@ -0,0 +1,70 @@
...
+    void signal(const char8* buffer, count_t count);
+    void wait() const;
+
+    const char8* data() const;
+    count_t count() const;
+
+    rsl::string_view filepath() const;
+
+  private:
+    rsl::string_view m_filepath;
```
Dyronix, line 40: Do we need to store the file path?
Is there a use for retrieving the file path?

source/1_engine/rex_engine/include/rex_engine/filesystem/mount_roots.h
```diff
@@ -7,7 +7,7 @@ namespace rex
 {
   enum class MountRoot
```
Dyronix, line 8: I think we can come up with a better name for this.
MountingPoint?

source/1_engine/rex_engine/include/rex_engine/filesystem/vfs.h
```diff
@@ -0,0 +1,70 @@
...
+    const char8* m_buffer;
+    count_t m_count;
+  };
+
+  void init(rsl::string_view root = "");
+  void mount(MountRoot root, rsl::string_view path);
+  void shutdown();
+
+  REX_NO_DISCARD rsl::unique_array<char8> open_read(rsl::string_view filepath);
+  REX_NO_DISCARD rsl::unique_array<char8> open_read(MountRoot root, rsl::string_view filepath);
```
Dyronix, line 52: memory::blob

source/1_engine/rex_engine/include/rex_engine/filesystem/vfs.h
```diff
@@ -0,0 +1,70 @@
...
+  REX_NO_DISCARD ReadRequest open_read_async(rsl::string_view filepath);
+  REX_NO_DISCARD ReadRequest open_read_async(MountRoot root, rsl::string_view filepath);
+  void save_to_file(rsl::string_view filepath, const void* data, card64 size, AppendToFile shouldAppend);
+  void save_to_file(MountRoot root, rsl::string_view filepath, const void* data, card64 size, AppendToFile shouldAppend);
```
Dyronix, line 56: memory::blob

source/1_engine/rex_engine/include/rex_engine/filesystem/vfs.h
```diff
@@ -0,0 +1,70 @@
...
+  bool exists(rsl::string_view path);
+  bool exists(MountRoot root, rsl::string_view path);
+  bool is_dir(rsl::string_view path);
+  bool is_dir(MountRoot root, rsl::string_view path);
+  bool is_file(rsl::string_view path);
+  bool is_file(MountRoot root, rsl::string_view path);
+  bool is_abs(rsl::string_view path);
+  bool is_rel(rsl::string_view path);
+  rsl::medium_stack_string create_full_path(rsl::string_view path);
+  rsl::medium_stack_string create_full_path(MountRoot root, rsl::string_view path);
```
Dyronix, line 69: I forgot how big a medium stack string is but if this cannot exceed OS max path length you can ignore this comment

source/1_engine/rex_engine/include/rex_engine/filesystem/vfs.h
```diff
@@ -0,0 +1,70 @@
+#pragma once
```
Dyronix, line 1: Going to put this here, but it might be an idea to use a "Path" structure ( even if it is a using for rsl::string_view ) this just give a better overview of what is expected as input for the VFS.

source/1_engine/rex_engine/rex_engine.sharpmake.cs
```diff
@@ -25,9 +26,11 @@ public RexEngine() : base()
     string relative_source_path = Util.PathGetRelative(Path.Combine(Globals.SourceRoot), SourceRootPath);
     GenerationConfigPath = Path.Combine(Globals.Root, "config", relative_source_path, "generation.json");
     MemoryTagsHeaderFile = Path.Combine(SourceRootPath, "include", "rex_engine", "memory", "memory_tags.h");
+    FileSystemMountHeaderFile = Path.Combine(SourceRootPath, "include", "rex_engine", "filesystem", "mount_roots.h");
```
Dyronix, line 29: Why do we need this?

Dyronix (changes requested):

source/1_engine/rex_engine/include/rex_engine/filesystem/mount_roots.h
RisingLiberty: What's the difference? It's the root path of a mount

source/1_engine/rex_engine/include/rex_engine/filesystem/vfs.h
RisingLiberty: It cannot

source/1_engine/rex_engine/include/rex_engine/filesystem/mount_roots.h
Dyronix: If it does not make any difference to you we can change this to MountingPoint
I find MountRoot a bit weird.
MountingPoint represents more as a location ( cuz it has point in the name ) in regards to MountRoot.

source/1_engine/rex_engine/include/rex_engine/filesystem/vfs.h
RisingLiberty: Yes, I want to use a path structure as well, it'll be using a string underneath though, not a string view

build/sharpmake/src/base.cs
RisingLiberty: it's some other boyscout changes that got into this change

build/sharpmake/src/base.cs
RisingLiberty: the memory tags didn't need to get updated, but the code that auto generates enums got updated

build/sharpmake/src/base.cs
Dyronix: An you refrain from doing no related stuff on a feature branch?

source/1_engine/rex_engine/include/rex_engine/filesystem/vfs.h
RisingLiberty: I can't use blob because a blob is supposed to be the owner of its memory, this is not

source/1_engine/rex_engine/include/rex_engine/filesystem/vfs.h
RisingLiberty: I can't use blob because a blob is supposed to be the owner of its memory, this is not

source/1_engine/rex_engine/include/rex_engine/filesystem/vfs.h
RisingLiberty: it's for debugging purposes

source/1_engine/rex_engine/rex_engine.sharpmake.cs
RisingLiberty: to avoid typos when deciding where memory goes.
this might seem pointless in a small framework like ours but useful when it grows bigger.
also makes it better to maintain if we ever decide to put data elsewhere (eg. test vs non-test)

source/1_engine/rex_engine/include/rex_engine/filesystem/vfs.h
Dyronix: Would it be possible to show how this is used?
Cuz you probably have an idea in your head but I cannot read your mind =D.

source/1_engine/rex_engine/include/rex_engine/filesystem/vfs.h
Dyronix: so it does make sense to have a blob_view or something in that manner, to store memory we do not own.

source/1_engine/rex_engine/include/rex_engine/filesystem/vfs.h
Dyronix: Made an issue for this

Dyronix (dismissed):

Dyronix:
Would it be possible to resolve the merge conflicts on this branch

Dyronix (approved):

PR #120, opened 2023-05-13
Development/input

RisingLiberty:


source/2_platform/rex_windows/include/rex_windows/console_application.h
```diff
@@ -0,0 +1,35 @@
...
+      bool platform_init() override;
+      void platform_update() override;
+      void platform_shutdown() override;
+
+    private:
+      class Internal;
+      static constexpr card32 s_internal_size      = 264;
+      static constexpr card32 s_internal_alignment = 8;
+      rsl::aligned_storage<s_internal_size, s_internal_alignment> m_internal_obj;
+      Internal* m_internal;
```
Dyronix, line 32: All of this is to have a PIMPL structure ... 
I would still go for a rsl::unique_ptr and be done with it, an application is only created once.
If this was a hot path and we would have a lot of these objects I would understand why you would do this but this seems overkill

source/2_platform/rex_windows/src/entrypoint.cpp
```diff
@@ -25,20 +28,38 @@ INT APIENTRY WinMain(HINSTANCE hInstance, HINSTANCE hPrevInstance, LPTSTR lpCmdL
...
+  }
 
   // by this point the application has finished and shutdown
   REX_LOG(LogWindows, "Application completed with result: {0}", result);
 
   return result;
 }
 
+// main is always the entry point.
+// on a graphical application however, we call into WinMain
```
Dyronix, line 56: We also call into win main on a none graphical application following this code.

source/2_platform/rex_windows/src/input/input.cpp
```diff
@@ -0,0 +1,72 @@
...
+namespace rex
+{
+  namespace input
+  {
+    // NOLINTBEGIN(cppcoreguidelines-avoid-non-const-global-variables)
+    card32 g_current_input_idx               = 1;
+    rsl::array<bool, 256> g_keys_even_frames = {};
+    rsl::array<bool, 256> g_keys_odd_frames  = {};
+    rsl::array<bool, 256>* g_prev_keys       = &g_keys_even_frames;
+    rsl::array<bool, 256>* g_current_keys    = &g_keys_odd_frames;
```
Dyronix, line 20: Why are these globals?

source/2_platform/rex_windows/src/input/input.cpp
```diff
@@ -0,0 +1,72 @@
...
+      g_current_keys->fill(false);
+
+      while(const uint8 key = input::poll())
+      {
+        if(key == '\0')
+          break;
+
+        g_current_keys->at(key) = true;
+      }
+    }
```
Dyronix, line 53: I have no idea how yet but I would try to shield this function from the client.
Maybe the same implementation we came up with for delta time ( if this was even applied ).
You don't want the client to be able to call this, client should only be able to retrieve the key state.

source/4_tools/console_app/src/console_app.cpp
```diff
@@ -0,0 +1,109 @@
...
+    // 
+    // Functionality should be implemented in memory_tracking.cpp
+    // However, we'll need a file system first to save things to disk
+    // We can then write a tool that scans this file for a memory analysis
+    // 
+    // Requirements of this tool:
+    // - Open a memory report file saved by the engine (See above)
+    // - Track all live allocations with their callstacks
+    // - Diff between different points in time
+    // - Track all single frame allocations
```
Dyronix, line 41: this explanation should not go into the repo, just make a ticket for it in the project overview of rex

Dyronix (changes requested):

source/2_platform/rex_windows/src/input/input.cpp
RisingLiberty: why shouldn't they be?

Dyronix: allocated memory in global scope instead of using a function for it and returning a static.

RisingLiberty: what is allocated?

RisingLiberty: there's no dynamic allocated memory here

Dyronix: No, but you still make room for 256 bytes (* 4) while if there is no input this is unused memory. Aren't you the one who mentioned keeping a low memory profile?

RisingLiberty: it doesn't matter if they're in a function or not, the memory will be there anyway, it'll be in the exe data sector

Dyronix: It loads the memory already even within a function?

RisingLiberty: yes, it's static memory, so it'll always be there

Dyronix: But if the function is not called how does it know it needs to initialize this?

RisingLiberty: initialization is different from memory allocation.
the memory is allocated but the lifetime of the object isn't started until the function is called for the first time

Dyronix: Where have you seen this?
Cuz I've googled it and I cannot find a reference, they only talk about initialization.

RisingLiberty: https://stackoverflow.com/questions/93039/where-are-static-variables-stored-in-c-and-c

Dyronix: Oke I understand now, there is no use in moving this.

source/2_platform/rex_windows/include/rex_windows/console_application.h
RisingLiberty: this is fixed in the console app PR

source/2_platform/rex_windows/src/entrypoint.cpp
RisingLiberty: a comment has been added that console applications also use winmain

source/2_platform/rex_windows/src/input/input.cpp
RisingLiberty: I put it in an internal namespace as we discussed

source/4_tools/console_app/src/console_app.cpp
RisingLiberty: the console app is not yet added to rex, I'm currently using it as a test app, so I don't think the above is a concern

RisingLiberty: it's not added to rex solution, just to the repo

Dyronix (approved):

Dyronix:
Would it be possible to resolve the merge conflicts of this branch

Dyronix (approved):

PR #124, opened 2023-05-14
[devops/clang_tidy] add enable all clang-tidy checks to generate script

RisingLiberty:


_generate.py
```diff
@@ -19,6 +19,7 @@
 if __name__ == "__main__":
   parser = argparse.ArgumentParser(formatter_class=argparse.RawTextHelpFormatter)
   parser.add_argument("-no_clang_tools", help="don't generate a compiler db", action="store_true")
+  parser.add_argument("-enable_all_checks", help="Enable all clang-tidy checks", action="store_true")
```
Dyronix, line 22: Are these all check that we defined or "All" checks?

RisingLiberty: No, asan, ubsan, .. they're not checking anything themselves, those compile the ececutable in a special way, but you still need to run the exe, clang-tidy is the one that has static analysis checks

Dyronix: I still don't understand what this does, what checks are enabled? Doesn't test.py -clang_tidy do all checks?

RisingLiberty: post_build.py by default only performs the first pass of checks, meaning the include reordering and namespace comment placements.
all checks means it'll run the same checks as test.py as a second pass as well.

RisingLiberty: yes, test.py does all checks, but that's irrelevant here, test.py doesn't use the frontend scripts (_generate.py, _build.py)
All this does is add a argument to the frontend script to tell the backend script to pass to sharpmake that it should enable all checks when post_build.py is called after a successful build

Dyronix: Can you add to the description of this argument that by default only reorder and namespace placements are enabled.

Dyronix (approved):

PR #128, opened 2023-05-15
[boyscout/improve_enum_generation] improved enum auto generation

RisingLiberty:


build/sharpmake/src/base.cs
```diff
@@ -254,6 +254,11 @@ public override void Configure(RexConfiguration conf, RexTarget target)
     }
```
Dyronix, line 254: Isn't this file becoming a bit bloated with all the different projects inside of it?

build/sharpmake/src/main.cs
```diff
@@ -291,6 +293,65 @@ private static void GenerateCompilerDatabases()
...
+    {
+      sb.AppendLine($"    // {className} values for {project_tags.Key}");
+
+      foreach (string tag in project_tags.Value)
+      {
+        sb.AppendLine($"    {tag},");
+      }
+    }
+
+    sb.AppendLine("  };");
```
Dyronix, line 322: Wouldn't it be better to use an ISourceGenerator here?
This would make the code a bit more readable I think.

https://learn.microsoft.com/en-us/dotnet/api/microsoft.codeanalysis.isourcegenerator?view=roslyn-dotnet-4.3.0
https://learn.microsoft.com/en-us/dotnet/csharp/roslyn-sdk/source-generators-overview

build/sharpmake/src/main.cs
```diff
@@ -291,6 +293,65 @@ private static void GenerateCompilerDatabases()
...
+    sb.AppendLine("");
+    sb.AppendLine("namespace rex");
+    sb.AppendLine("{");
+  }
+
+  private static void WriteCustomGenerationFooter(StringBuilder sb)
+  {
+    sb.AppendLine("} // namespace rex");
+    sb.AppendLine("// NOLINTEND");
+  }
```
Dyronix, line 343: Perhaps these lines can also be included in the ISourceGenerator

Dyronix (changes requested):

build/sharpmake/src/base.cs
RisingLiberty: maybe, but that shouldn't be part of this change.
no extra projects got added in this change

build/sharpmake/src/main.cs
RisingLiberty: from a quick read source generator is to generate C# code not C++ code

Dyronix: You are correct, my bad.
But I still think this part of the code can become a bit more readable we should not apply it within this PR but maybe come up with something to generate C++ code. 

This is probably not the last code that we will generate, will be useful to have a system for that.

Dyronix (approved):

Dyronix:
Would it be possible to resolve the merge conflicts on this branch please?

Dyronix (approved):

PR #129, opened 2023-05-15
[boyscout/adapter_redesign] redesigned adapter system to avoid the down cast

RisingLiberty:


source/2_platform/rex_directx/src/directx_renderer.cpp
```diff
@@ -80,11 +80,12 @@ namespace rex
         }
 
         const dxgi::AdapterManager adapter_manager(&factory, &internal::highest_scoring_gpu);
-        const dxgi::Adapter* adapter = static_cast<const dxgi::Adapter*>(adapter_manager.selected()); // NNOLINT(cppcoreguidelines-pro-type-static-cast-downcast)
+        const Gpu* selected_gpu = adapter_manager.selected();
+        IDXGIAdapter* adapter = static_cast<IDXGIAdapter*>(selected_gpu->native_handle());
```
Dyronix, line 84: I like the idea where you are going with but for somebody who told me he wants to remove the void* on from the rendering context ( gui_app.cpp - context::create ) not sure if this is the approach we should go for.

You still need a cast.

Dyronix (changes requested):

Dyronix (approved):

Dyronix (approved):

PR #130, opened 2023-05-16
Devops/include what you use

RisingLiberty:


source/iwyu.imp
```diff
@@ -0,0 +1,377 @@
+[
```
Dyronix, line 1: What does this file do?

Dyronix (changes requested):

source/iwyu.imp
RisingLiberty: it maps headers from internal headers to headers that should be used
see https://github.com/include-what-you-use/include-what-you-use/blob/master/docs/IWYUMappings.md

Dyronix (approved):

Dyronix:
Would it be possible to resolve the merge conflicts on this branch please?

Issue #68, opened 2023-04-12
Investigate unused code within rexlog.


Dyronix:
Certain sinks will never be used in the near future.
It is better to remove these.
- qt_sink
- android_sink
- ... ?

PR #133, opened 2023-05-21
Development/diagnostics remove unused code

Dyronix:
https://github.com/New-Creta/Rex/issues/68


source/1_engine/rex_engine/include/rex_engine/function_signature_define.h
```diff
@@ -0,0 +1,19 @@
+#pragma once
```
RisingLiberty, line 1: don't think this file should be in the root of rex_engine.
I assume this is copy pasted from spdlog? can it be cleaned up a bit?

we don't use compiler macros directly, we use REX_COMPILER_MSVC etc, please use those macros.
it's well defined which version of clang (15.0.0) and msvc (14.29.30133) we use.

I'd base the code around that instead of the below as it's not really clear what we're using

afaik __PRETTY_FUNCTION__  for anything but MSVC and __FUNCSIG__ for MSVC is fine nowadays

source/1_engine/rex_engine/src/diagnostics/logging/logger_config.cpp
```diff
@@ -114,11 +121,24 @@ namespace rexlog
...
+        rexlog::details::Registry::instance().set_levels(rsl::move(levels), global_level_found ? &global_level : nullptr);
       }
 
     } // namespace helpers
-  }   // namespace cfg
-} // namespace rexlog
 
-// NOLINTEND(misc-definitions-in-headers)
+    // search for REXLOG_LEVEL= in the args and use it to init the levels
+    void load_log_levels(const rsl::vector<rsl::string_view>& args)
```
RisingLiberty, line 130: I'd like to have a good way of getting all the possible commandline arguments for the engine, like you can do with python "--help" command.
it's outside of the scope of this change, but just to let you know that this code will change slightly to follow that design

source/1_engine/rex_engine/src/diagnostics/logging/logger_config.cpp
```diff
@@ -114,11 +121,24 @@ namespace rexlog
...
+    {
+      const rex::DebugString rexlog_level_prefix("REXLOG_LEVEL=");
```
RisingLiberty, line 132: please use a string_view here

source/1_engine/rex_engine/src/diagnostics/logging/logger_config.cpp
```diff
@@ -114,11 +121,24 @@ namespace rexlog
...
+
+      for(rsl::string_view s : args)
+      {
+        if(s.find(rexlog_level_prefix) == 0)
```
RisingLiberty, line 136: note: we do have rsl::string_type::contains but in this case not sure if you should use it, as it could introduce a bug if the cli doesn't start with "REXLOG_LEVEL="

source/1_engine/rex_engine/src/diagnostics/logging/logger_config.cpp
```diff
@@ -114,11 +121,24 @@ namespace rexlog
...
+        {
+          auto levels_string = s.substr(rexlog_level_prefix.size());
+          helpers::load_levels(rex::DebugString(levels_string));
```
RisingLiberty, line 139: the interface of load_levels, doesn't have to take a rsl::DebugString, a string_view would work just fine, and then you can avoid a possible allocation here

source/1_engine/rex_engine/src/diagnostics/logging/logger_config.cpp
```diff
@@ -89,17 +96,17 @@ namespace rexlog
...
-        rex::DebugHashTable<rex::DebugString, level::LevelEnum> levels;
-        level::LevelEnum global_level = level::Info;
+        rex::DebugHashTable<rex::DebugString, rexlog::level::LevelEnum> levels;
+        rexlog::level::LevelEnum global_level = rexlog::level::Info;
         bool global_level_found       = false;
 
         for(auto& name_level: key_vals)
         {
           const auto& logger_name = name_level.key;
           auto level_name         = rex::DebugString(to_lower_impl(name_level.value));
```
RisingLiberty, line 106: I don't think this should be an debug string here. it returns itself

source/2_platform/rex_windows/src/entrypoint.cpp
```diff
@@ -17,6 +17,11 @@
 INT APIENTRY WinMain(HINSTANCE hInstance, HINSTANCE hPrevInstance, LPTSTR lpCmdLine, int nShowCmd)
 {
   rex::CommandLineArguments cmd_args(GetCommandLineA());
+
+#if REX_DEBUG
+  rex::diagnostics::load_log_levels(cmd_args.arguments());
```
RisingLiberty, line 22: this would be ideal for pre_app entry as well, when that change goes into main

RisingLiberty (changes requested):
small changes required, more info in comments

RisingLiberty:
also please fix CI, clang-tidy failed

source/1_engine/rex_engine/src/diagnostics/logging/logger_config.cpp
Dyronix: How am I supposed to create a DebugStringStream from a rsl::string_view?

PR #151, opened 2023-05-25
Development/command line ags

RisingLiberty:


source/1_engine/rex_engine/src/cmdline.cpp
```diff
@@ -0,0 +1,178 @@
...
+    private:
+      rsl::vector<ActiveArgument> m_arguments;
+    };
+
+    rsl::unique_ptr<CommandLineArguments> g_cmd_line_args; // NOLINT(fuchsia-statically-constructed-objects, cppcoreguidelines-avoid-non-const-global-variables)
+
+    void init(rsl::string_view cmdLine)
+    {
+      g_cmd_line_args = rsl::make_unique<CommandLineArguments>(cmdLine);
+    }
```
Dyronix, line 146: Shouldn't this be the pattern we always use?
Static in a function?

source/1_engine/rex_engine/src/cmdline.cpp
```diff
@@ -0,0 +1,178 @@
...
+              return false;
+            }
+          }
+        }
+
+        return true;
+      }
+
+    private:
+      rsl::vector<ActiveArgument> m_arguments;
```
Dyronix, line 138: It's it going to work better if we use a hash?
We will look for command line arguments more often than inserting them.

source/1_engine/rex_engine/src/cmdline.cpp
```diff
@@ -0,0 +1,178 @@
...
+        auto cmd_it = rsl::find_if(g_command_line_args.cbegin(), g_command_line_args.cend(), [key](const CommandLineArgument& cmdArg) { return key == cmdArg.name; });
+
+        if(cmd_it == g_command_line_args.cend())
+        {
+          REX_WARN(LogEngine, "Command {} passed in but it's not recognised as a valid command so will be ignored", arg);
+          return;
+        }
+
+        m_arguments.push_back({key, value});
+      }
```
Dyronix, line 113: So what will happen if I have multiple entire for the same argument.
For instance?

-libs_to_compile=spdlog+crcpp+libpng

Are we going to parse this defining the separator of values internally
Is the client going to parse this?

source/1_engine/rex_engine/src/filesystem/vfs.cpp
```diff
@@ -311,6 +286,42 @@ namespace rex
...
+    void init(rsl::string_view root)
+    {
+      if(rsl::optional<rsl::string_view> arg = cmdline::get_argument("Root"))
+      {
+        root = arg.value();
+      }
+
+      if(root.empty())
+      {
+        WIN_CALL(GetCurrentDirectoryA(g_root.max_size(), g_root.data()));
```
Dyronix, line 298: This should not be within the VFS directly.
This should be shielded behind a platform layer

source/1_engine/rex_engine/src/filesystem/vfs.cpp
```diff
@@ -311,6 +286,42 @@ namespace rex
...
+        g_root.reset_null_termination_offset();
+      }
+      else if(is_abs(root))
+      {
+        g_root = root;
+      }
+      else
+      {
+        rsl::medium_stack_string cwd;
+        WIN_CALL(GetCurrentDirectoryA(cwd.max_size(), cwd.data()));
```
Dyronix, line 308: Same here

source/1_engine/rex_engine/src/filesystem/vfs.cpp
```diff
@@ -311,6 +286,42 @@ namespace rex
...
+        cwd.reset_null_termination_offset();
+
+        g_root = cwd;
+        g_root += "/";
+        g_root += root;
```
Dyronix, line 313: Can't we use a more platform agnostic approach here?
path::combine?

source/2_platform/rex_windows/src/gui_application.cpp
```diff
@@ -128,14 +127,21 @@ namespace rex
 
       void subscribe_window_events()
       {
-        event_system::subscribe(event_system::EventType::WindowClose, [this](const event_system::Event& /*evt*/) { m_app_instance->quit(); });
+        event_system::subscribe(event_system::EventType::WindowClose,
+                                [this](const event_system::Event& /*evt*/)
+                                {
+                                  rex::event_system::Event ev {};
+                                  ev.type = rex::event_system::EventType::QuitApp;
+                                  rex::event_system::fire_event(ev);
```
Dyronix, line 135: This is a direct call IIRC?
You should be careful when calling events from another event.
The event_system should queue events and poll for available events ( maybe even a max per frame ) 
No part of this PR but we should create an issue for it.

source/4_tools/regina/src/regina.cpp
```diff
@@ -40,8 +40,14 @@ namespace rex
...
     return app_params;
   }
+
+#ifndef REX_ENABLE_AUTO_TESTS
+
+  ApplicationCreationParams app_entry(PlatformCreationParams&& platformParams)
+  {
+    return create_regina_app_creation_params(rsl::move(platformParams));
+  }
+#endif
```
Dyronix, line 52: We should come up with a solution to get rid of this macro

Dyronix (changes requested):

source/1_engine/rex_engine/src/filesystem/vfs.cpp
RisingLiberty: not part of this PR I'd say

source/1_engine/rex_engine/src/filesystem/vfs.cpp
RisingLiberty: not part of this PR I'd say

source/1_engine/rex_engine/src/filesystem/vfs.cpp
RisingLiberty: not part of this PR I'd say

source/2_platform/rex_windows/src/gui_application.cpp
RisingLiberty: I think events should be processed on a different tick than when they're added
events added on tick n should be processed on tick n + 1 to avoid any issues
I don't thin we should limit the number of events we can process in a frame though, maybe just warn if it's too many

source/4_tools/regina/src/regina.cpp
RisingLiberty: Regina can have a start project which checks at runtime if autotests are specified and then calls into Regina or ReginaTest, then you don't need a macro, you do need an extra project though, one that depends on both Regina and ReginaTest

source/1_engine/rex_engine/src/filesystem/vfs.cpp
Dyronix: No, but still has to be fixed IMO, this is a platform call that should not be here.
Although this is added code on this branch, maybe it came from a merge into this branch.

source/2_platform/rex_windows/src/gui_application.cpp
Dyronix: Well, that's what I'm saying right queue them up.

source/1_engine/rex_engine/src/cmdline.cpp
RisingLiberty: can't do this as it needs to get initialized with an argument

source/1_engine/rex_engine/src/cmdline.cpp
RisingLiberty: yes, but we'll have to hash it every time we want to look it up with kinda defeats the purpose don't you think?

source/1_engine/rex_engine/src/cmdline.cpp
RisingLiberty: the client is supposed to parse this.
cmdline is only associating keys with parse (would break if you pass the same argument multiple times though, so that needs to be caught)

source/1_engine/rex_engine/src/cmdline.cpp
RisingLiberty: what we could do, once compile time sid are in, is pass in an sid and check that.
that'll be hashed at compile time so no runtime overhead

Dyronix (dismissed):

Dyronix:
Can you fix CI please?

source/1_engine/rex_engine/include/rex_engine/diagnostics/logging/logger_config.h
```diff
@@ -6,6 +6,6 @@ namespace rex
 {
   namespace diagnostics
   {
-    void load_log_levels(const rsl::vector<rsl::string_view>& args);
+    void init();
```
Dyronix, line 9: I don't think we should adjust the name of this function ... init does not tell me a lot. Load log levels explains what the function does.

source/1_engine/rex_engine/src/diagnostics/logging/logger_config.cpp
```diff
@@ -28,81 +32,74 @@ namespace rex
...
       {
         if(input.empty() || input.size() > 512)
         {
           return;
         }
 
-        auto key_vals = extract_key_vals_impl(input);
-        rex::DebugHashTable<rex::DebugString, rexlog::level::LevelEnum> levels;
+        auto key_vals = extract_key_vals(input);
+        rsl::unordered_map<rsl::string_view, rexlog::level::LevelEnum> levels;
```
Dyronix, line 95: You made a "using" for this, we should use it

source/1_engine/rex_engine/include/rex_engine/diagnostics/logging/internal/details/registry.h
```diff
@@ -30,7 +30,7 @@ namespace rexlog
     class Registry
     {
     public:
-      using LogLevels = rex::DebugHashTable<rex::DebugString, level::LevelEnum>;
+      using LogLevels = rsl::unordered_map<rsl::string_view, level::LevelEnum>;
```
Dyronix, line 33: Aren't these stored, so making it a string_view seems kinda dangerous to me

source/1_engine/rex_engine/src/entrypoint.cpp
```diff
@@ -1,12 +1,51 @@
...
+      // we'll have a timer in place to break for 2 minutes, if no debugger is attached
+      // we close down the program
+      if (cmdline::get_argument("BreakOnBoot"))
+      {
+        using namespace rsl::chrono_literals;
+        auto i = 1s;
+        while (i < 10min)
+        {
+          rsl::this_thread::sleep_for(1s);
+          ++i;
```
Dyronix, line 39: I would encapsulate the timing functionality, as this might become useful elsewhere

Dyronix (changes requested):

source/1_engine/rex_engine/src/diagnostics/logging/logger_config.cpp
RisingLiberty: I made using for one using a debug allocator, this one shouldn't

source/1_engine/rex_engine/src/entrypoint.cpp
RisingLiberty: Where else do you see this exact code useful?
If you want to wait, you often sleep for x seconds
I can't do that here as it's possible a debugger gets attached

source/1_engine/rex_engine/include/rex_engine/diagnostics/logging/logger_config.h
RisingLiberty: But it's not loading them, it's initialising them.

source/1_engine/rex_engine/include/rex_engine/diagnostics/logging/internal/details/registry.h
RisingLiberty: they come directly from the commandline, if none are passed in it uses the global level instead

source/1_engine/rex_engine/include/rex_engine/diagnostics/logging/logger_config.h
Dyronix: Then you might call it initialize log levels, diagnostics::init() is a bbit generic

source/1_engine/rex_engine/src/diagnostics/logging/logger_config.cpp
Dyronix: I cannot check this on my phone, but IIRC there is a using for this somewhere, could you double check pls

source/1_engine/rex_engine/include/rex_engine/diagnostics/logging/internal/details/registry.h
Dyronix: Btw isn't this the same type

RisingLiberty: what's the difference?

RisingLiberty: if no log levels are passed on the command line, the global one is used

source/1_engine/rex_engine/src/entrypoint.cpp
Dyronix: For instance, AI might not update every frame. Objects in the distance might run at 12 fps rather than 60 so there could be some update controller which manages this that waits for an update

RisingLiberty: yes, that's fine, but that shouldn't follow the above code, they wait for x seconds in 1 go, they don't iterate on the amount of time they're waiting.

RisingLiberty: also I'd think such a system has a dependency on the engine being initialized and this update controller being up and running etc.
here I want to have a simple timer, allow someone to attach a debugger and immediately continue executing without having to wait for the remaining time to finish.
if no debugger gets attached in 10min, I want the app to close or it'd keep running forever.
it's important almost nothing else gets done before it reaches this point or it defeats the purpose of it

source/1_engine/rex_engine/include/rex_engine/diagnostics/logging/internal/details/registry.h
Dyronix: I mean isn't this the same type as the using I mentioned before

RisingLiberty: no, well this is the using you're talking about I think

source/1_engine/rex_engine/src/entrypoint.cpp
Dyronix: Okay, true and we cannot create a callback ( a promise ) that is called when x amount of time is passed so this implementation can become a bit more generic?

RisingLiberty: we can, but it'd have to look the same as above which I don't think we're gonna need.
this is basically "spinlock", it checks something, if not true yet, sleeps and repeats this n times.
if we want to generalise this, that's fine, but that's outside of the scope of this PR I'd say

source/1_engine/rex_engine/src/diagnostics/logging/logger_config.cpp
RisingLiberty: fixed it

source/1_engine/rex_engine/src/entrypoint.cpp
Dyronix: We'll discuss this later it's fine for now ;-)

Dyronix (approved):

Dyronix (approved):

PR #153, opened 2023-05-25
Boyscout/cleanup code generation

RisingLiberty:


_build/sharpmake/src/code_generaton.cs
```diff
@@ -1,34 +1,66 @@
+using System.Collections.Generic;
 using System.Text.Json;
+using System.IO;
 using Sharpmake;
+using System.Text;
 
 public class CodeGeneration
```
Dyronix, line 7: Mark the class as static as there should never be an instance of this class

Dyronix (approved):

PR #166, opened 2023-05-27
[bug/app_update_when_paused] app can update when paused, the client d…

RisingLiberty:
…ecides what happens (as not everything should get updated when the app is paused

### DIFF
--- a/_build/config/required_externals.json
+++ b/_build/config/required_externals.json
@@ -2,7 +2,7 @@
   "rex_std": {
     "name": "rex_std",
     "url": "https://github.com/RisingLiberty/RexStl/",
-    "tag": "v1.1.42",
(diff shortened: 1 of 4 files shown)


Dyronix:
I would not approach it like this, update can just be called ( even when the application is paused ) we just skip render logic.

Dyronix:
As discussed, just skip the rendering code and have the application update ( make sure this does not eat all resources though -> the framecap should still apply ), later we can add a frame cap for a paused application so the update loop does not run full force when paused.

Dyronix:
In fact maybe the framecap is already possible, as we already have one implemented if the application is paused just lower the max fps

Dyronix (approved):

PR #177, opened 2023-05-29
Development/tests

RisingLiberty:


config/tests/regina_test/generation.json
```diff
@@ -0,0 +1,14 @@
+{
```
Dyronix, line 1: generated?

source/1_engine/rex_engine/include/rex_engine/diagnostics/logging/logger_config.h
```diff
@@ -6,6 +6,6 @@ namespace rex
 {
   namespace diagnostics
   {
-    void load_log_levels(const rsl::vector<rsl::string_view>& args);
+    void init();
```
Dyronix, line 9: This is fixed in the other branch right?

source/1_engine/rex_engine/src/cmdline.cpp
```diff
@@ -0,0 +1,205 @@
...
+{
+  namespace cmdline
+  {
+    struct ActiveArgument
+    {
+      rsl::string_view argument;
+      rsl::string_view value;
+    };
+
+    class CommandLineArguments
```
Dyronix, line 20: Just a question, but since we change the other class "CommandLineArgument" to cmdline::Argument. shouldn't we call this class cmdline::Arguments? Although I can see why we should keep it as this as this might become confusing with the cmdline::Argument class. One might think this is a list of this "Argument" class which it is not.

source/1_engine/rex_engine/src/cmdline.cpp
```diff
@@ -0,0 +1,205 @@
...
+    {
+    public:
+      explicit CommandLineArguments(rsl::string_view cmdLine)
+      {
+        // verify the auto generated command line arguments
+        REX_ASSERT_X(verify_args(g_command_line_args.data(), g_command_line_args.size()), "You have ambuguous command line arguments");
```
Dyronix, line 26: Not sure if this should be an assert, we will be running auto tests in release and probably have release versions of the engine as well. there might still be some "ambiguous" arguments in release which would result in issues in release/auto tests. I think and error or a REX_VERIFY_X ( asserts in release ) might be in order here.

source/1_engine/rex_engine/src/cmdline.cpp
```diff
@@ -0,0 +1,205 @@
...
+        parse_cmd_line(rsl::string_view(cmdLine));
+      }
+
+      rsl::optional<rsl::string_view> get_argument(rsl::string_view arg)
+      {
+        for(ActiveArgument active_arg: m_arguments)
+        {
+          // early optimization that strincmp can't do
+          if(arg.length() != active_arg.argument.length())
+            continue;
```
Dyronix, line 43: Add { }

source/1_engine/rex_engine/src/cmdline.cpp
```diff
@@ -0,0 +1,205 @@
...
+          REX_LOG(LogEngine, "No command line arguments to parse");
+          return;
+        }
+
+        parse_cmd_line(rsl::string_view(cmdLine));
+      }
+
+      rsl::optional<rsl::string_view> get_argument(rsl::string_view arg)
+      {
+        for(ActiveArgument active_arg: m_arguments)
```
Dyronix, line 39: Is this an intentional copy?

source/1_engine/rex_engine/src/cmdline.cpp
```diff
@@ -0,0 +1,205 @@
...
+    private:
+      void parse_cmd_line(rsl::string_view cmdLine)
+      {
+        // all we have to do is split the command line based with spaces
+        // there's no need to perform any heap allocation for that.
+        count_t start_pos = cmdLine.find_first_not_of(' ', 0); // it's possible, although unlikely, that we start with a space
+
+        // all executables should be called with at least 1 argument
+        // the first argument should always be the path to the executable itself
+        REX_ASSERT_X(start_pos != -1, "No arguments given to your exe, there should always be at least 1 arg.");
```
Dyronix, line 63: Maybe also REX_VERIFY_X although less error prone? As this might get caught in a debug session first.

source/1_engine/rex_engine/src/cmdline.cpp
```diff
@@ -0,0 +1,205 @@
...
+        {
+          return;
+        }
+
+        start_pos         = space_pos + 1;
+
+        // get the new space pos
+        space_pos = cmdLine.find_first_of(' ', start_pos);
+
+        const rsl::string_view arg_prefix = "-"; // all arguments should start with a '-'
```
Dyronix, line 81: What if they don't?

source/1_engine/rex_engine/src/cmdline.cpp
```diff
@@ -0,0 +1,205 @@
...
+          const rsl::string_view argument = cmdLine.substr(start_pos + arg_prefix.size(), length);
+          add_argument(argument);
+          start_pos = cmdLine.find_first_not_of(' ', space_pos); // skip all additional spaces
+          space_pos = cmdLine.find_first_of(' ', start_pos);
+        }
+
+        if(start_pos != -1)
+        {
+          add_argument(cmdLine.substr(start_pos + arg_prefix.size())); // + 1 to ignore '-'
+        }
```
Dyronix, line 94: Dunno if I asked this already or though about it. But what with multiple values for a single key.
eg: -Config=debug+release

source/1_engine/rex_engine/src/cmdline.cpp
```diff
@@ -0,0 +1,205 @@
...
+        {
+          REX_WARN(LogEngine, "Command {} passed in but it's not recognised as a valid command so will be ignored", arg);
+          return;
+        }
+
+        auto active_it = rsl::find_if(m_arguments.cbegin(), m_arguments.cend(), [key](const ActiveArgument& activeArg) { return activeArg.argument == key; });
+
+        if(active_it != m_arguments.cend())
+        {
+          REX_WARN(LogEngine, "Command {} was already passed in. passing the same argument multiple times is not supported. will be skipped", key);
```
Dyronix, line 128: Should we maybe return the invalid one such as in Python? Not sure if it's useful though.

source/1_engine/rex_engine/src/cmdline.cpp
```diff
@@ -0,0 +1,205 @@
...
+        {
+          key   = arg;
+          value = "1"; // this is so we can easily convert to bool/int
+        }
+
+        auto cmd_it = rsl::find_if(g_command_line_args.cbegin(), g_command_line_args.cend(), [key](const Argument& cmdArg) { return key == cmdArg.name; });
+
+        if(cmd_it == g_command_line_args.cend())
+        {
+          REX_WARN(LogEngine, "Command {} passed in but it's not recognised as a valid command so will be ignored", arg);
```
Dyronix, line 120: Should we maybe return the invalid one such as in Python? Not sure if it's useful though.

source/1_engine/rex_engine/src/cmdline.cpp
```diff
@@ -0,0 +1,205 @@
...
+
+      bool verify_args(const Argument* args, count_t argCount) // NOLINT(readability-convert-member-functions-to-static)
+      {
+        for(count_t i = 0; i < argCount; ++i)
+        {
+          const Argument& lhs_arg = args[i];
+          for(count_t j = 0; j < argCount; ++j)
+          {
+            if(i == j)
+              continue;
```
Dyronix, line 143: Add "{ }"

source/1_engine/rex_engine/src/cmdline.cpp
```diff
@@ -0,0 +1,205 @@
...
+      g_cmd_line_args = rsl::make_unique<CommandLineArguments>(cmdLine);
+    }
+
+    void print_args()
+    {
+      REX_LOG(LogEngine, "Listing Command line arguments");
+
+      rsl::unordered_map<rsl::string_view, rsl::vector<Argument>> project_to_arguments;
+
+      for(Argument cmd: g_command_line_args)
```
Dyronix, line 174: Intentional copy?

source/1_engine/rex_engine/src/cmdline.cpp
```diff
@@ -0,0 +1,205 @@
...
+      {
+        project_to_arguments[cmd.module].push_back(cmd);
+      }
+
+      for(auto [project, cmds]: project_to_arguments)
```
Dyronix, line 179: Not really a fan of structured bindings ... dunno how you feel about it but I rather have a struct.
But you can keep it.

source/1_engine/rex_engine/src/cmdline.cpp
```diff
@@ -0,0 +1,205 @@
...
+      {
+        REX_LOG(LogEngine, "");
+        REX_LOG(LogEngine, "Commandline Arguments For {}", project);
+        REX_LOG(LogEngine, "------------------------------");
+
+        for(Argument cmd: cmds)
```
Dyronix, line 185: I get the idea this is intentional

source/1_engine/rex_engine/src/filesystem/vfs.cpp
```diff
@@ -1,5 +1,6 @@
 #include "rex_engine/filesystem/vfs.h"
```
Dyronix, line 1: I think the first auto tests should be written for this class ...
Even just to test the API

source/1_engine/rex_engine/src/filesystem/vfs.cpp
```diff
@@ -311,6 +286,42 @@ namespace rex
...
+    void init(rsl::string_view root)
+    {
+      if(rsl::optional<rsl::string_view> arg = cmdline::get_argument("Root"))
+      {
+        root = arg.value();
+      }
+
+      if(root.empty())
+      {
+        WIN_CALL(GetCurrentDirectoryA(g_root.max_size(), g_root.data()));
```
Dyronix, line 298: This still has to move, but I have seen an issue for this

source/1_engine/rex_engine/src/filesystem/vfs.cpp
```diff
@@ -311,6 +286,42 @@ namespace rex
...
+        g_root.reset_null_termination_offset();
+      }
+      else if(is_abs(root))
+      {
+        g_root = root;
+      }
+      else
+      {
+        rsl::medium_stack_string cwd;
+        WIN_CALL(GetCurrentDirectoryA(cwd.max_size(), cwd.data()));
```
Dyronix, line 308: Same for this one

source/1_engine/rex_engine/src/filesystem/vfs.cpp
```diff
@@ -311,6 +286,42 @@ namespace rex
...
+        cwd.reset_null_termination_offset();
+
+        g_root = cwd;
+        g_root += "/";
```
Dyronix, line 312: We should use the REX_FILE_SEPARATOR ( part of the rexlog refactor )

source/1_engine/rex_engine/src/filesystem/vfs.cpp
```diff
@@ -311,6 +286,42 @@ namespace rex
...
+        g_root += root;
+      }
+
+      g_is_initialized  = true;
+      g_keep_processing = true;
```
Dyronix, line 317: Maybe bit overkill but we have a statecontroller we could utilize it here.

source/2_platform/rex_windows/src/gui_application.cpp
```diff
@@ -126,14 +125,21 @@ namespace rex
...
       void subscribe_window_events()
       {
-        event_system::subscribe(event_system::EventType::WindowClose, [this](const event_system::Event& /*evt*/) { m_app_instance->quit(); });
+        event_system::subscribe(event_system::EventType::WindowClose,
+                                [this](const event_system::Event& /*evt*/)
+                                {
+                                  rex::event_system::Event ev {};
+                                  ev.type = rex::event_system::EventType::QuitApp;
+                                  rex::event_system::fire_event(ev);
+                                });
```
Dyronix, line 134: Just as a reminder dunno if this is already in place but we should queue up events not call them directly this is dangerous

Dyronix (changes requested):

_build/sharpmake/src/base.cs
```diff
@@ -519,4 +469,23 @@ public override void Configure(RexConfiguration conf, RexTarget target)
...
+    base.Configure(conf, target);
+
+    if (GenerateSettings.AddressSanitizerEnabled)
+    {
+      conf.add_public_define("CATCH_CONFIG_DISABLE"); // we don't need to check catch, it massively increase link time (47min at time of writing -> 5min)
+    }
+    else if (GenerateSettings.UndefinedBehaviorSanitizerEnabled)
+    {
+      conf.add_public_define("CATCH_CONFIG_DISABLE"); // we don't need to check catch, it massively increase link time (47min at time of writing -> 5min)
+    }
```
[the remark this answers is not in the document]
RisingLiberty: asan or ubsan? yes

_build/sharpmake/src/generate_settings.cs
```diff
@@ -13,12 +13,32 @@ public class EnumGenerationConfig
   public string Filepath { get; set; }
 }
 
+public class ArrayGenerationConfig
+{
+  public string ElementType { get; set; }
+  public string Name { get; set; }
+  public List<string> Includes { get; set; }
+  public List<string> Values { get; set; }
+  public string Filepath { get; set; }
```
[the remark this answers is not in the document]
RisingLiberty: I don't know what you mean?

_build/sharpmake/src/generate_settings.cs
```diff
@@ -13,12 +13,32 @@ public class EnumGenerationConfig
...
+}
+
 // this is the object we use
 public class EnumGenerationSettings
 {
   public string ClassName = "";
   public string Filepath = "";
```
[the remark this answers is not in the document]
RisingLiberty: I believe they're null, but this is for readability

_build/sharpmake/src/globals.cs
```diff
@@ -11,7 +11,7 @@ public class BuildSettings
   public string[] misc_folders { get; set; }
   public string[] misc_extensions { get; set; }
   public string build_folder { get; set; }
-  public string test_folder { get; set; }
+  public string tests_folder { get; set; }
   public string tools_folder { get; set; }
   public string libs_folder { get; set; }
   public string source_folder { get; set; }
```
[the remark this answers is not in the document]
RisingLiberty: not sure what you mean?

_build/sharpmake/src/main.cs
```diff
@@ -68,7 +68,8 @@ public virtual void Configure(RexConfiguration conf, RexTarget target)
     {
       if (file.EndsWith(".sharpmake.cs"))
       {
-        sharpmake_sources.Add($"{quote}{file}{quote}");
+        string fileWithSlashes = file.Replace('\\', '/');
```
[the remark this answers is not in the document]
RisingLiberty: the path separator is \\ on windows which is what I'm trying to avoid here

_test.py
```diff
@@ -53,7 +53,7 @@
     if args.auto_test_timeout:
       auto_test_timeout_secs = args.auto_test_timeout
 
-    regis.test.run_auto_tests(["debug", "debug_opt", "release"], ["msvc","clang"], ["regina"], int(auto_test_timeout_secs), args.clean, args.single_threaded)
+    regis.test.run_auto_tests(["debug", "debug_opt", "release"], ["msvc","clang"], ["reginatest"], int(auto_test_timeout_secs), args.clean, args.single_threaded)
```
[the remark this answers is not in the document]
RisingLiberty: regina test is testing regina.
before it launched regina and closed it down within 10 seconds as a quick autotest, now we have proper auto tests so we use those to auto test the editor.

source/1_engine/rex_engine/include/rex_engine/diagnostics/logging/logger_config.h
RisingLiberty: yes

source/1_engine/rex_engine/src/cmdline.cpp
RisingLiberty: yea, well I think this class can just be renamed to something like cmdline::ArgumentsProcessor or something

source/1_engine/rex_engine/src/cmdline.cpp
RisingLiberty: it doesn't process ambiguous arguments and will always process the first one, that's why this is an assert here.
I don't think it matters if we have a release version of the engine or not, if you have 2 arguments in the list, eg 1 coming from the engine, 1 coming from the client, the engine doesn't know which one you want to use, which is why there's an assert here.

source/1_engine/rex_engine/src/cmdline.cpp
RisingLiberty: no, I don't think so

config/source/1_engine/rex_engine/generation.json
```diff
@@ -0,0 +1,38 @@
+{
```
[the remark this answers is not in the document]
RisingLiberty: this file is not auto generated, it's used for the auto generation

config/source/1_engine/rex_renderer_core/generation.json
```diff
@@ -0,0 +1,12 @@
+{
```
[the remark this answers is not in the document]
RisingLiberty: no

config/source/2_platform/rex_opengl/generation.json
```diff
@@ -0,0 +1,12 @@
+{
```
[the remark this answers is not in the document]
RisingLiberty: no

config/source/2_platform/rex_windows/generation.json
```diff
@@ -0,0 +1,12 @@
+{
```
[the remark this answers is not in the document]
RisingLiberty: no

config/source/4_tools/console_app/generation.json
```diff
@@ -1,9 +1,12 @@
 {
```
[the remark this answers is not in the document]
RisingLiberty: no

config/source/4_tools/regina/generation.json
```diff
@@ -0,0 +1,12 @@
+{
```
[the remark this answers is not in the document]
RisingLiberty: no

config/tests/regina_test/generation.json
RisingLiberty: no

source/1_engine/rex_engine/src/cmdline.cpp
RisingLiberty: I'm not sure what we'd win if we have a REX_VERIFY here?

source/1_engine/rex_engine/src/cmdline.cpp
RisingLiberty: then you get an error that your argument isn't found

source/1_engine/rex_engine/src/cmdline.cpp
RisingLiberty: that's fine, in the logger_config.cpp you see how that's processed
arguments are seperated by spaces, in the example above, the value of the argument would be debug+release, the client is responsible for parsing that

source/1_engine/rex_engine/src/cmdline.cpp
RisingLiberty: not sure what we gain with that, we just ignore the next iteration and print a warning about it

source/1_engine/rex_engine/src/cmdline.cpp
RisingLiberty: not sure what we gain with that, we just ignore it and print a warning about it

source/1_engine/rex_engine/src/cmdline.cpp
RisingLiberty: it's not intentional, it's just copying 2 string_views, it's not that big of a concern which is why I didn't capture it by const ref

source/1_engine/rex_engine/src/filesystem/vfs.cpp
RisingLiberty: yes, I agree, but outside of this PR.
this PR is just implementing autotests and adds 1 as an example how the API should be used

source/1_engine/rex_engine/src/filesystem/vfs.cpp
RisingLiberty: we should use Path.join here, when that class is added

source/1_engine/rex_engine/src/filesystem/vfs.cpp
RisingLiberty: these are just flags, the state controller is more to have multiple states at once (and used with an enum) so I think it's a bit overkill to use it here

source/2_platform/rex_windows/src/gui_application.cpp
RisingLiberty: do we have an issue for this?
the code in here shouldn't change imo, other than cleaning up how to fire an event cause it's 3 lines of code now while it should just be 1.
on backend site it should put this in a different queue than the one it's currently processing

_build/sharpmake/src/generate_settings.cs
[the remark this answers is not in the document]
Dyronix: You can just write 

public my_property;

_build/sharpmake/src/globals.cs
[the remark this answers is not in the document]
Dyronix: See previous comment

source/1_engine/rex_engine/src/cmdline.cpp
Dyronix: Yes but the problem is that this is only caught in debug ... not sure if this scenario will ever occur in debug

source/1_engine/rex_engine/src/cmdline.cpp
Dyronix: Yeah my point exactly, I'm just a little concerned of potential issues ( as in programming mistakes and not user input ) within the auto tests as asserts are only fired within debug and most of our tests should run in release.

source/1_engine/rex_engine/src/cmdline.cpp
Dyronix: Perhaps some utility class might be created to extract values giving in a custom separator. Not for this PR though

source/1_engine/rex_engine/src/cmdline.cpp
Dyronix: Yeah, I was just thinking out loud

source/1_engine/rex_engine/src/cmdline.cpp
RisingLiberty: It can in debug but if you're comment means to also check this in release, that's fine

source/1_engine/rex_engine/src/cmdline.cpp
RisingLiberty: All of our tests should run in all configs imo

Dyronix: Not sure if that's useful, tests make sure functionality is not broken in other places and final functionality is release no? So I'm not sure if we should waste resources ( as in hardware resources ) to test in debug as this might give a misguided view on things

source/1_engine/rex_engine/src/filesystem/vfs.cpp
Dyronix: Is there an issue for this?

source/2_platform/rex_windows/src/gui_application.cpp
Dyronix: Is there an issue for the queue?

source/1_engine/rex_engine/src/cmdline.cpp
RisingLiberty: We should become more than once something is broken in 1 config and not in others

RisingLiberty: But we can have release only tests

source/1_engine/rex_engine/src/filesystem/vfs.cpp
RisingLiberty: Yes

source/2_platform/rex_windows/src/gui_application.cpp
RisingLiberty: Yes

source/1_engine/rex_engine/src/filesystem/vfs.cpp
Dyronix: Create an issue for this and this conversation is also resolved

source/1_engine/rex_engine/src/cmdline.cpp
Dyronix: Keep consistency with other structures, a copy is not needed here. Two string views is 24 bytes no? Unless the size is stored in an s64 them it's 32 bytes ( this is not aligned ( 24 ) without padding IIRC )

RisingLiberty: It is alligned afaik, but regardless, I converted it to const ref

source/1_engine/rex_engine/src/cmdline.cpp
Dyronix: But debug is caught while we develop ... but sure if you are convinced we need it. I just don't see a point in when this could benefit us. Unless we start developing cross platform, debug runs will just result in what we run locally

source/1_engine/rex_engine/src/cmdline.cpp
Dyronix: Yes, but I was wondering if there might be more and if so I don't wanna repeat code for checking an assert in release so that is why I proposed a REX_VERIFY

source/1_engine/rex_engine/src/cmdline.cpp
RisingLiberty: not all things are caught while we develop, I already found issues in the boot tests that we didn't know about

source/1_engine/rex_engine/src/cmdline.cpp
RisingLiberty: you can have a assert in release or something else, but then it's not really an assert (or it doesn't have to be behind a macro anyway)
it's a runtime check that's always there, regardless of the configuration you run

Dyronix (approved):

PR #180, opened 2023-05-30
Development/stringpool api

Dyronix:


source/1_engine/rex_engine/src/string/stringpool.cpp
```diff
@@ -52,25 +52,25 @@ namespace rex
       auto result = get_entries().emplace(rsl::move(entry_id), rsl::move(entry));
       if(result.emplace_successful)
       {
-        return &result.inserted_element->key;
+        return result.inserted_element->key;
       }
 
       REX_ASSERT("This path should never be reached, insertion into the string pool failed somehow.");
-      return nullptr;
+      return {};
```
RisingLiberty, line 59: can this be named instead or return StringEntryID(invalid_something_something);
eg. a global constexpr invalid stringid (like std::optional has with std::nullopt)

source/1_engine/rex_engine/src/string/stringpool.cpp
```diff
@@ -80,11 +80,11 @@ namespace rex
         REX_ASSERT_X(it != rsl::cend(get_entries()), "StringID::is_none() not present");
       }
 
-      return &(it->value);
+      return (it->value);
```
RisingLiberty, line 83: unnecessary brackets

RisingLiberty (changes requested):

source/1_engine/rex_engine/src/string/stringpool.cpp
Dyronix: This has to become an invalid_object but we don't have the code available on this branch.

RisingLiberty (changes requested):
Please resolve the merge conflicts

RisingLiberty (approved):

PR #183, opened 2023-05-30
Development/compile time stringid

Dyronix:


source/1_engine/rex_engine/include/rex_engine/string/stringid.h
```diff
@@ -11,54 +11,56 @@
 
 namespace rex
 {
-  class StringID
-  {
-  public:
-    StringID();
-    explicit StringID(rsl::string_view characters);
+    class StringViewID;
```
RisingLiberty, line 14: I don't think this is compile time, nothing is made constexpr
the StringID needs to make every member a constexpr function for it to be usable at compile time.
same goes for the string view id.

also where do you see StringViewIDs being used where a StringID would be a bad choice?
I'm guessing you want to use this on a literal instead of a string_view to save on time when doing comparisons?

RisingLiberty (changes requested):

Dyronix:
This branch can only succeed if the following PR is merged into stable/RSL: https://github.com/RisingLiberty/RexStl/pull/3
Version of RSL should be bumped on this branch as well to be able to retrieve the changes

RisingLiberty (approved):

PR #211, opened 2023-06-05
Development/replace stringentryid with stringid

Dyronix:


source/1_engine/rex_engine/src/string/stringid.cpp
```diff
@@ -2,19 +2,21 @@
...
   {
     //-------------------------------------------------------------------------
-    StringEntryID make_and_store(rsl::string_view characters)
+    StringID make_and_store(rsl::string_view characters)
     {
       if(characters.empty())
       {
         REX_WARN(LogEngine, "Trying to create a StringID with an empty string, this is not allowed.");
-        return StringEntryID(); // StringID::is_none() == true
+        return StringID(); // StringID::is_none() == true
```
RisingLiberty, line 19: this can use StringID::create_invalid() now, no?

RisingLiberty (changes requested):
just 1 comment

PR #212, opened 2023-06-05
[development/rex_defines] Extract definition that might be utilized s…

Dyronix:
…omewhere else into a different library ( or header file )


source/1_engine/rex_engine/include/rex_engine/filesystem/path_defines.h
```diff
@@ -0,0 +1,16 @@
+#pragma once
+
+#ifndef REX_PATH_DEFINES
+  #define REX_PATH_DEFINES
+
+  //-------------------------------------------------------------------------
+  // File path separator
+  #ifndef REX_FOLDER_SEPS
+    #ifdef REX_PLATFORM_WINDOWS
+      #define REX_FOLDER_SEPS "\\/"
```
RisingLiberty, line 10: can this be converted to global, inline constexpr string_views?

source/1_engine/rex_engine/include/rex_engine/defines.h
```diff
@@ -29,6 +29,16 @@
     #error DEBUG_BREAK unsupported machine instruction ...
   #endif
 
+  //-------------------------------------------------------------------------
+  // End of line
+  #ifndef REX_EOL
+    #ifdef REX_PLATFORM_WINDOWS
+      #define REX_EOL "\r\n"
```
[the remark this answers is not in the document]
RisingLiberty: I know they're used to init a string_view, but I don't know why we don't branch on platform there instead.
we want to avoid using defines as global constants as much as possible

source/1_engine/rex_engine/include/rex_engine/filesystem/path_defines.h
RisingLiberty: I know they're used to init a string_view, but I don't know why we don't branch on platform there instead.
we want to avoid using defines as global constants as much as possible

source/2_platform/rex_windows/src/entrypoint.cpp
```diff
@@ -14,22 +14,22 @@
 #include <iostream>
 #include <shellapi.h>
 
-#pragma warning (disable : 4702)
+#pragma warning(disable : 4702)
```
RisingLiberty, line 17: can you please remove this (it came from some debugging I did).

RisingLiberty (changes requested):
please have a look at the review comments

source/1_engine/rex_engine/include/rex_engine/defines.h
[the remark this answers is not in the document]
Dyronix: You want to move the global constants as well and use those instead of the defines?

source/1_engine/rex_engine/include/rex_engine/filesystem/path_defines.h
Dyronix: Same comment ( You want to move the global constants as well and use those instead of the defines? )

source/1_engine/rex_engine/include/rex_engine/defines.h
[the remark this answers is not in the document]
RisingLiberty: yea, I would've gone into that direction, I don't see a gain in splitting it up into a define and a global constant really
other than moving the branching to a different place

source/1_engine/rex_engine/include/rex_engine/defines.h
```diff
@@ -29,6 +31,19 @@
     #error DEBUG_BREAK unsupported machine instruction ...
   #endif
 
+//-------------------------------------------------------------------------
+// End of line
+namespace rex
```
RisingLiberty, line 36: shouldn't really be in defines.h anymore now though..
but I'll make an issue for this so we can clean it up another time (I can take into my cleanup branch when I go over the repo)

RisingLiberty (dismissed):

RisingLiberty (approved):

Issue #205, opened 2023-06-04
create REX_VERIFY


RisingLiberty:
Create REX_VERIFY which equals an assert that's also valid in release

PR #215, opened 2023-06-19
Development/rex verify

Dyronix:


source/1_engine/rex_engine/include/rex_engine/diagnostics/verify.h
```diff
@@ -0,0 +1,20 @@
...
+  rex::rex_verify(rsl::format(__VA_ARGS__));                                                                                                                                                                                                             \
+  while(true)                                                                                                                                                                                                                                            \
+  {                                                                                                                                                                                                                                                      \
+    DEBUG_BREAK();                                                                                                                                                                                                                                       \
```
RisingLiberty, line 14: this looks to be a copy of the assert implementation with just an extra thing to open a dialog
can't it use the existing assert implementation and show the messagebox afterwards

_build/config/required_externals.json
```diff
@@ -1,20 +1,26 @@
...
+        "tag": "v1.0.0",
+        "storage": "~/source/0_thirdparty"
+    },
+    "rex_glm": {
+        "name": "rex_glm",
+        "url": "https://gitlab.com/rex-externals/rex-glm/",
+        "tag": "v1.0.0",
+        "storage": "~/source/0_thirdparty"
+    },
+    "rex_tiny_file_dialogs": {
```
RisingLiberty, line 20: I actually wanted to write this myself
I'm guessing it's a cross platform library to show message boxes etc?

RisingLiberty (changes requested):

source/1_engine/rex_engine/include/rex_engine/diagnostics/verify.h
RisingLiberty: the library seems pretty small so maybe we'll port it into the engine

RisingLiberty:
so you've created a verify, which is an assert in release, but I don't see it used anywhere
what are the rules when to use a verify and when to use an assert?

source/1_engine/rex_engine/include/rex_engine/diagnostics/verify.h
Dyronix: I could yes but that looked a bit confusing to me to be honest as we created verify to have a check in release cuz asserts are compiled out but then we still call assert code in release. I can reuse the code but it felt contradicting. 

I don't wanna fall into the spdlog trap again with this.
By reinventing the wheel again and implementing this ourselves.

_build/config/required_externals.json
Dyronix: I don't see any added value in writing this ourselves.
Yes, it's a cross-platform implementation of a message box.

Dyronix:
> so you've created a verify, which is an assert in release, but I don't see it used anywhere what are the rules when to use a verify and when to use an assert?

It will be used on the code you've written once it's reviewed and merged. There was a comment for this.
I would say we use "verify" for things that should not fail no matter what, such as graphics initialization or VFS setup, the engine cannot start if those are not available. "Assert" is a programmer mistake being checked, where verify seems to me like a more critical issue.

source/1_engine/rex_engine/include/rex_engine/diagnostics/verify.h
RisingLiberty: yes, it shouldn't use be called "rex_assert" then, you can rename the func to something more generic and use that instead. can put a comment there that it's used by both assert and verify

_build/config/required_externals.json
RisingLiberty: the added value of writing it yourself is so that it can use internal rex code and be a bit more performant possibly, depends how it's implemented

RisingLiberty:
asserts are also for critical mistakes, asserts are for any mistake where the code cannot recover from it.
can you remind what the comment was for the verify?

Dyronix:
> asserts are also for critical mistakes, asserts are for any mistake where the code cannot recover from it. can you remind what the comment was for the verify?

IMO asserts are different. 

Or is a check if something is nullptr and not being able to initialize the graphics api the same level of error to you? 

I use asserts to make sure API is used correctly and there are no easy to catch issues within the flow of the application ( such as nullptr for instance ). 

But a verify is a check on a bigger system such as "renderer failed to initialize" "window could not be created",

_build/config/required_externals.json
RisingLiberty: you're using this in verify which is a assert for release which is not development, so messageboxes will apear in non development builds as well.
I can also think of other cases where a messagebox will be helpful that an end user would like to see when something wrong happens

RisingLiberty: it's non critical code, for sure and I'm sure it's not that slow, but the more it's integrated into our engine code, the better, it'll just take more time

RisingLiberty: question is how much time we want to spend on it

RisingLiberty:
this is to resolve https://github.com/New-Creta/Rex/issues/205

RisingLiberty:
I've finished the implementation, it can be reviewed now

Dyronix:
"This basically acts like a wrapper around a condition, if the condition fails a msg box is prompted to the user indicating something went wrong"

What is the difference between having an assert which is also a wrapper around a condition. difference is it logs instead of popping a message box.

RisingLiberty:
The difference is that asserts are compiled out, verification is not

Issue #202, opened 2023-06-04
Create a second event queue, one is processed, the other for new events


RisingLiberty:
Is there an issue for the queue?

_Originally posted by @Dyronix in https://github.com/New-Creta/Rex/pull/177#discussion_r1216705808_

PR #216, opened 2023-06-19
Development/event queue

Dyronix:


source/1_engine/rex_engine/src/event_system.cpp
```diff
@@ -14,18 +14,75 @@ namespace rex
...
+            ,m_front()
+            ,m_back()
+            ,m_current(&m_front)
+            ,m_next(&m_back)
+        {
+        }
+
+        void subscribe(EventType type, const EventFunction& function)
+        {
+          m_subscriptions[type].push_back(function);
```
RisingLiberty, line 33: just to note, this is not thread save, as in, the entire class is not thread safe

source/1_engine/rex_engine/src/event_system.cpp
```diff
@@ -14,18 +14,75 @@ namespace rex
...
+              delegate(evt);
+            }
+          }
+
+          m_current->clear();
+        }
+
+        void present()
+        {
+          rsl::vector<Event>* temp = m_current;
```
RisingLiberty, line 58: why not use rsl::swap?

RisingLiberty (changes requested):

source/1_engine/rex_engine/src/event_system.cpp
Dyronix: I think there are a lot more functionalities within REX that are not thread-safe. As we are single-threaded for now I think it's fine.

source/1_engine/rex_engine/src/event_system.cpp
Dyronix: Dunno, it does the same thing just within a function call.
I'll adjust it if you want.

RisingLiberty: that's why we have functions, to wrap logic so I'd recommend to use it here

RisingLiberty:
this is to resolve https://github.com/New-Creta/Rex/issues/202

RisingLiberty:
I updated the code, it's ready for review

RisingLiberty:
I think the best way to have this have multiple producer single consumer kinda thing.
multiple threads can push events but only 1 thread processes them.
that's currently not implemented because we don't have a thread pool/job system yet

RisingLiberty (approved):

PR #226, opened 2023-12-15
Development/rex fuzzy tests

RisingLiberty:
Created a dummy fuzzy testing project for rex engine.


_build/scripts/test.py
```diff
@@ -25,6 +25,7 @@
   parser.add_argument("-single_threaded", help="run tests in single threaded mode", action="store_true")
   parser.add_argument("-only_errors_and_warnings", help="filter lines to only display warnings and errors", action="store_true")
   parser.add_argument("-auto_fix", help="auto fix where you can (applies to iwyu and clang-tidy)", action="store_true")
+  parser.add_argument("-runs", help="num runs to perform (applies to fuzzy testing)", default=10000)
```
Dyronix, line 28: I think I already commented this somewhere else but can we split up the different type of tests in a new subprocess like we do with setup/generate/build. This will solve that we have arguments for specific runs within a general place.

_build/sharpmake/src/base.cs
```diff
@@ -53,14 +53,26 @@ private void SetupDefaultConfigurationSettings(RexConfiguration conf, RexTarget
     conf.Options.Add(Options.Vc.General.CharacterSet.MultiByte);
     conf.Options.Add(Options.Vc.General.PlatformToolset.v142);
     conf.Options.Add(Options.Vc.General.WarningLevel.Level4);
-    conf.Options.Add(Options.Vc.General.TreatWarningsAsErrors.Enable);
+
+    // Disable warning as errors in debug so that we can add debug code
+    // without the compiler warning about it.
+    // We just need to make sure we delete it before it goes in to version control
```
Dyronix, line 59: INFO: this will be captured by the CI when it's still present. CI will also run Release and this will trigger the warning as error.

Thanks for implementing the feature request.

_build/sharpmake/src/main.cs
```diff
@@ -184,7 +184,7 @@ private static void InitializeGenerationSettings(Dictionary<string, ConfigSettin
 
     ProjectGen.Settings.ClangTidyRegex = config["clang-tidy-regex"].Value.GetString();
     ProjectGen.Settings.PerformAllClangTidyChecks = config["perform-all-clang-tidy-checks"].Value.GetBoolean();
-    ProjectGen.Settings.NoClangTools = config["no-clang-tools"].Value.GetBoolean();
+    ProjectGen.Settings.ClangToolsEnabled = config["no-clang-tools"].Value.GetBoolean() == false;
```
Dyronix, line 187: INFO: Is there a reason why this remains "no-clang-tools" in the config file?

Dyronix (changes requested):
I was wondering if there is an option to disable all the testing. Most CMAKE projects have this option to disable the tests (locally ofc) this will only clutter the sln (if you use Visual Studio) for instance: when you are not working on REX but building on top of REX, like when you start a new UE5 project.

_build/scripts/test.py
RisingLiberty: you want to be able to do something like
`py _rex.py fuzz` `py _rex.py unit-test` `py _rex.py auto-test` `py _rex.py static-analysis`?
that's doable, although, probably best in another PR

_build/sharpmake/src/base.cs
RisingLiberty: true, the warning as error will be triggered, but ideally you want this at submit time..
we can perform a compilation of the files we've changed only on when you submit them to capture these issues.
you really want to capture these things as early as possible. another way of doing it is by manually tagging these lines and let the submit step parse a file looking for tags which would block submission

_build/scripts/test.py
Dyronix: Yes that is what I want, another PR is fine

_build/sharpmake/src/base.cs
Dyronix: I would not complicate things with extra tags
You can trigger a release compilation before you commit, but this will prevent us to push code if it does not compile ( which is maybe what you want ) but it will slow down development if you would like to push code from one place to another or give somebody else control of the code if it does not compile.

_build/sharpmake/src/main.cs
Dyronix: Would we have to add this flag to disable the clang tools every time we request a new generation cuz then I rather not but otherwise I would make it consistent with the rest

RisingLiberty: I'll disable clang tools by default, that'll make things easier

_build/scripts/test.py
RisingLiberty: linked to https://github.com/New-Creta/Rex/issues/272

_build/sharpmake/src/main.cs
RisingLiberty: done

Dyronix (approved):

PR #227, opened 2023-12-15
Development/static analysis through python

RisingLiberty:
Added a python script that scan for clang-tidy ignored warnings and std usage in the codebase

### DIFF
--- a/_build/scripts/static_analysis/static_analysis.py
+++ b/_build/scripts/static_analysis/static_analysis.py
@@ -51,7 +51,6 @@ def scan_clang_tidy_ignore(sourceRoots : list[str], excludeFolders : list[str]):
     ignored_warnings = {}
 
     for found_match in found_matches:
-        # regis.diagnostics.log_debug(f'{found_match.filepath}({found_match.line_nr}): {found_match.line}')
         ignore_text = found_match.line[found_match.start_pos:]
         start = ignore_text.find('(')+1
         end = ignore_text.find(')')
--- a/_build/scripts/static_analysis/static_analysis.py
+++ b/_build/scripts/static_analysis/static_analysis.py
@@ -1,0 +1,92 @@
+import os
+import re
+import regis.util
+import regis.diagnostics
+
+source_root_folders = [
+    'source',
+    'tests'
+]
+
+exclude_folders = [
+    '0_thirdparty',
+    'catch2'
+]
+   
+class FoundMatch():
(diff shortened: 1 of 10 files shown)


_build/scripts/static_analysis/static_analysis.py
```diff
@@ -0,0 +1,92 @@
+import os
```
RisingLiberty, line 1: This is the only file that needs review

_build/scripts/static_analysis/static_analysis.py
```diff
@@ -0,0 +1,92 @@
...
+    regis.diagnostics.log_info(f'scanning for clang-tidy ignore messages')
+    regex_pattern = r'\/\/\s*NOLINT\(' # scanning for "// NOLINT("
+    found_matches : list[FoundMatch] = []
+    for source_root in sourceRoots:
+        found_matches.extend(scan_root(regex_pattern, source_root, excludeFolders))
+
+    ignored_warnings = {}
+
+    for found_match in found_matches:
+        # regis.diagnostics.log_debug(f'{found_match.filepath}({found_match.line_nr}): {found_match.line}')
```
Dyronix, line 54: Please don't push commented code

_build/scripts/static_analysis/static_analysis.py
```diff
@@ -0,0 +1,92 @@
...
+        regis.diagnostics.log_info(f'{found_match.filepath}({found_match.line_nr}): {found_match.line.rstrip()}')
+
+def log_header(msg : str):
+    regis.diagnostics.log_info('------------------------')
+    regis.diagnostics.log_info(msg)
+    regis.diagnostics.log_info('------------------------')
+
+def main():
+    log_header(f'scanning for clang-tidy ignores')
+    scan_clang_tidy_ignore(source_root_folders, exclude_folders)
```
Dyronix, line 87: INFO: Why do we need this is this just for logging purposes that we know what is getting ignored?

Dyronix (changes requested):

_build/scripts/static_analysis/static_analysis.py
RisingLiberty: I plan on adding it to CI in a way, maybe automatically submit it into version control so we can see what we ignore and where etc, if we ever decide to change our clang-tidy flags, we can base it on that data

Dyronix (approved):

PR #228, opened 2023-12-16
Docs/test projects

RisingLiberty:
Added documentation on how to write all the different kind of test projects.
- unit tests
- auto tests
- fuzzy tests


Dyronix:
I have nowhere to put this so I will add it as a comment here.
Would it be possible to add a feature that is able to update a single external without updating the file in required_externals?
IIRC this is also possible within Unreal. 

for instance: _rex.py setup libs=glad+glm -v=latest

_docs/src/documentation/rex_pipeline.md
```diff
@@ -89,11 +89,11 @@ Also known as combi builds and jumbo builds.
 
 Note: this has nothing to do with the Unity game engine.
 
-A unity build allows the build system to merge source files together, creating 1 big source files. This is simply done with C++ include statements including different .cpp files into intermediate .cpp files. 
+A unity build allows the build system to merge source files together, creating 1 big source file. This is simply done with C++ include statements including different .cpp files into intermediate .cpp files. 
 
 These intermediate files are then passed on to the compiler which improves compilation times a lot as the compiler doesn't have to open as many files and can parse more files in 1 go.
 
-Note: unity build support is only for Visual Studio generated projects, it's currently not supported for Ninja builds
+Note: Files modified on disk, different from their repo equivalent are excluded from unity files to improve compilation. They also have optimizations disabled so it's easier to debug them.
```
Dyronix, line 96: INFO: Does this mean that the code that I change locally is not included within the unity build? For instance I was developing the renderer so all the code that was included with that feature is not included within the unity build?

_docs/src/documentation/rex_pipeline.md
```diff
@@ -105,40 +105,70 @@ It'll scan for all sharpmake files recursively in 3 different folders
...
+
+There's only 1 extra commandline argument supported by the generate script which is `-sharpmake_args`. This is to allow a user to pass in arguments that are supported by the native sharpmake executable
+
+We support Visual Studio and Visual Studio Code as IDEs. These just act as text editors as they're in turn call into the build pipeline and build the ninja files which are always generated.
 
 Some Examples (Windows):
 ```sh
 # Default generation. generate files for engine and editor
-py _generate.py
+py _rex.py generate
```
Dyronix, line 120: Maybe show here what a default "generate" does what settings are enabled by default

_docs/src/documentation/rex_pipeline.md
```diff
@@ -105,40 +105,70 @@ It'll scan for all sharpmake files recursively in 3 different folders
...
+# Default generation + unit tests + code coverage + Visual Studio sln and projects
+py _rex.py generate -enable-unit-tests -enable-code-coverage -IDE VisualStudio
 
-# Default generation + disable clang tools. this skips the argument in the script and passes it directly to sharpmake 
-py _generate.py -sharpmake_args="/noClangTools" 
+# Default generation, but single threaded
+py _rex.py generate -sharpmake_args /multithreaded(false)
 ```
 
+### Code Generation
```
Dyronix, line 135: Can you show an example of this as well, is this a separate command, is it done automatically when you call generate?

_docs/src/documentation/rex_pipeline.md
```diff
@@ -105,40 +105,70 @@ It'll scan for all sharpmake files recursively in 3 different folders
...
+### Build Types
+We have a few different kind of build types aka configurations, some are only supported with Clang as they are sanitizers.
+The different builds types are
 
-the script takes the project, compiler and configuration you want to compile for.
-After which it looks for this the ninja project reflecting the project you specified (which is auto generated by sharpmake) and compiles the configuration specified with the compiler you specified.
+- Debug: No optimizations enabled. All debug information is available.
+- Debug Opt: Optimized version of Debug build.
+- Release: All optimizations enabled, no debug information is available.
+- Coverage: Only enabled for Clang. Enables code coverage reports
```
Dyronix, line 168: I would also explain code coverage and where those reports are located when generated.

_docs/src/documentation/rex_pipeline.md
```diff
@@ -105,40 +105,70 @@ It'll scan for all sharpmake files recursively in 3 different folders
...
+- [Address Sanitizer](https://clang.llvm.org/docs/AddressSanitizer.html): Only enabled for Clang. Used to detect out of bounds reads/writes. Double frees, use after free.
+- [Undefined Behavior Sanitizer](https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html): Only enabled for Clang. Used to detect undefined behavior at runtime.
+- [Fuzzy](https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html): Only enabled for Clang. Used for fuzzing projects
 
 ### Post Build
```
Dyronix, line 173: INFO: Can a user add it's own post_build events here?

_docs/src/documentation/rex_pipeline.md
```diff
@@ -148,16 +178,16 @@ This script launches clang-tidy and clang-format if clang tools are enabled. Bec
 Some Examples:
 ```sh
 # Builds regina project and its dependencies in debug mode using msvc compiler
-py _build.py -project=regina -config=debug -comp=msvc
+py _rex.py build -project=regina -config=debug -compiler=msvc
 
 # Cleans all regina and its depedencies' intermediate files and then performs a build afterwards
-py _build.py -project=regina -config=debug -comp=msvc -clean
+py _rex.py build -project=regina -config=debug -compiler=msvc -clean
 ```
```
Dyronix, line 185: These seem to be examples on how to trigger a build and have nothing to do with a post build event ( except that it gets triggered? )

_docs/src/documentation/rex_pipeline.md
```diff
@@ -148,16 +178,16 @@ This script launches clang-tidy and clang-format if clang tools are enabled. Bec
...
 
 ## Testing
-Rex Engine supports many different kind of tests for optimal testing which can all be fired from [`_test.py`](../../../_test.py).
+Rex Engine supports many different kind of tests for optimal testing which can all be fired from the root script
```
Dyronix, line 188: This has now changed into different sub-processes

Dyronix:
Also nowhere to comment this:

I think there is a typo here, shouldn't it be unit_tests instead of unity_tests
image

_docs/src/documentation/test_projects.md
```diff
@@ -0,0 +1,348 @@
+# How To Create Test Projects
+
+## Type Of Test Project
+
+Rex supports 3 different kind of test projects, each requiring a slightly different setup.
+
+### Unit Test Project
+A unit test project is ideal to test a library or system that doesn't need to engine to be in a specific state. If you have a container, allocator, algorithm, .. that you'd like to test, a unit test is ideal to do this.
+
+Let's work from an example. Let's pretend we want to create a unit test project for the rex engine project called `RexUnitTests`.
```
Dyronix, line 10: It would be a nice feature if there would be an automated step to setup these projects. for instance within VS they can just do File > New Project > Rex Unit Test Template and all the code is generated and put in place. It might also be a command line argument like we already do for enums and array we could also support generation of projects.

_docs/src/documentation/test_projects.md
```diff
@@ -0,0 +1,348 @@
...
+- We need to add a dependency to `RexUnitTest` project. Override the `SetupLibDependencies` func and add a public dependency to `RexUnitTest`.
+
+An example of such a file could look like this
+
+```cs
+using Sharpmake;
+using System.IO;
+
+[Generate]
+public class RexEngineUnitTests : TestProject
```
Dyronix, line 33: Not sure if there is any documentation yet about these base projects users can inherit from but that would be ideal. Lot of boiler plate needs to happen as well seems like this could benefit from some automation IMO. I think most unit tests projects would look very similar

_docs/src/documentation/test_projects.md
```diff
@@ -0,0 +1,348 @@
...
+
+    conf.Options.Remove(Options.Vc.Compiler.JumboBuild.Enable);
+  }
+}
+```
+
+Next you need to create a source file which will act as the entry point of the auto tests, let's call this file `regina_auto_tests.cpp`
+Fill the file with the following code
+
+```cpp
```
Dyronix, line 147: This seems like a lot more boiler plate isn't there any way to make this more user friendly I suppose most of the files would look quite similar. Isn't it possible to create a separate file that exposes the functions a user need and all the includes and basic setup are just within some other header file?

_docs/src/documentation/test_projects.md
```diff
@@ -0,0 +1,348 @@
...
+    return rex::auto_test::invalid_app_params(rsl::move(platformParams));
+  }
+} // namespace rex
+
+```
+
+This will add the boilerplate code that's needed to start the auto tests.
+We added a boot auto test example here, let's dig in to how you'd implement that.
+
+Create a file called `regina_boot_test.cpp`
```
Dyronix, line 207: A it seems like this is already the case ( this separate file ) why does the user need to create the previous file manually?

_docs/src/documentation/test_projects.md
```diff
@@ -0,0 +1,348 @@
...
+    return app_params;
+  }
+} // namespace regina_auto_test
+
+```
+
+That's all there is to it, once you have this code, you can run it and the editor boot will now be tested.
+If any crashes occur this will be reported by the testing pipeline
+
+### Fuzzy Test Project
```
Dyronix, line 260: Similar commets to Fuzzy tests

data/RexEngineFuzzyTest/corpus/dummy_corpus.txt
```diff
@@ -0,0 +1 @@
+this is a corpus file
```
Dyronix, line 1: INFO: What is a corpus file?

Dyronix (changes requested):

PR #230, opened 2023-12-18
Development/fuzzy boyscout

RisingLiberty:
Updating rexpy to the version that reads the corpus files and added a dummy corpus file to be read by the rex engine fuzzer

### DIFF
--- a/tests/1_engine/rex_engine_fuzzy_test/src/rex_engine_fuzzy_test.cpp
+++ b/tests/1_engine/rex_engine_fuzzy_test/src/rex_engine_fuzzy_test.cpp
@@ -6,10 +6,11 @@ namespace rex
 {
   namespace fuzzy
   {
-    int fuzzy_entry(fuzz_span input)
+    int fuzzy_entry(const fuzz_span& input)
     {
-
-      int test_val = rex::fuzzy::make_fuzzy_object<int>(input);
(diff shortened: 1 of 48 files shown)


tests/1_engine/rex_engine_fuzzy_test/src/rex_engine_fuzzy_test.cpp
```diff
@@ -6,10 +6,11 @@ namespace rex
...
   namespace fuzzy
   {
-    int fuzzy_entry(fuzz_span input)
+    int fuzzy_entry(const fuzz_span& input)
     {
-
-      int test_val = rex::fuzzy::make_fuzzy_object<int>(input);
+      // Example code
+      fuzz_span my_test_input = input;
+      int test_val = make_fuzzy_object<int>(my_test_input);
```
Dyronix, line 13: INFO: Does this need to remain here?

Dyronix (changes requested):

tests/1_engine/rex_engine_fuzzy_test/src/rex_engine_fuzzy_test.cpp
RisingLiberty: It's gone in one of the next PRs

PR #232, opened 2023-12-19
Development/adding crash pipeline entry

RisingLiberty:
This doesn't do much other than enabling structured exception handling. 
This allows us to perform custom logic when a crash occurs at runtime (eg. creating dumps, logs, ..)
This is useful for debugging and is especially useful for auto tests.

I haven't setup much, just the entry for this, I want to wait until all PRs have gone in before we start extending this or reviews will just stack up otherwise

### DIFF
--- a/source/2_platform/rex_windows/src/entrypoint.cpp
+++ b/source/2_platform/rex_windows/src/entrypoint.cpp
@@ -8,9 +8,19 @@
 #include "rex_windows/log.h"
 #include "rex_windows/platform_creation_params.h"
 
+#include "rex_std/iostream.h"
+
 #define NOMINMAX
 #include <Windows.h>
 #include <processenv.h>
+
+//-------------------------------------------------------------------------
+s32 report_crash(LPEXCEPTION_POINTERS ExceptionInfo)
+{
+  rsl::cout << "We crashed!\n";
+  (void)ExceptionInfo;
+  return EXCEPTION_EXECUTE_HANDLER;
+}
 
 //-------------------------------------------------------------------------
 int rex_entry(HINSTANCE hInstance, HINSTANCE hPrevInstance, LPTSTR lpCmdLine, int nShowCmd)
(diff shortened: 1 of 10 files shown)


source/2_platform/rex_windows/src/entrypoint.cpp
```diff
@@ -8,10 +8,20 @@
...
+#include "rex_std/iostream.h"
+
 #define NOMINMAX
 #include <Windows.h>
 #include <processenv.h>
 
+//-------------------------------------------------------------------------
+s32 report_crash(LPEXCEPTION_POINTERS ExceptionInfo)
+{
+  rsl::cout << "We crashed!\n";
```
Dyronix, line 20: If would use _set_se_translator here ( https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/set-se-translator?view=msvc-170 ) to receive a more meaningfull description.

source/2_platform/rex_windows/src/entrypoint.cpp
```diff
@@ -22,24 +32,37 @@ int rex_entry(HINSTANCE hInstance, HINSTANCE hPrevInstance, LPTSTR lpCmdLine, in
...
-  {
-    // this doesn't initialize anything but simply prepares the application for initialization
-    rex::win32::GuiApplication application(rsl::move(app_params));
 
-    // this initializes, runs the loop and performs the shutdown
-    result = application.run();
+  // We're wrapping executing in __try - __except block so that we can capture crashes
+  // This allows us to perform particular logic when we encounter a crash.
+  // Especially for when running tests, this can be very useful
+  __try
```
Dyronix, line 40: I think it wasn't recommended to use this, but I think there isn't really an alternative.

Dyronix (changes requested):

source/2_platform/rex_windows/src/entrypoint.cpp
RisingLiberty: my implementation is based on how Unreal does it, you can get the same info using the link above as how I'm doing it, one is the C way, the other the C++ way I think.
We discussed the this implementation is fine

PR #241, opened 2023-12-24
Development/file reorganisation

RisingLiberty:
Applied the platform file structure for files


source/2_platform/rex_windows/src/console_application.cpp
```diff
@@ -71,6 +72,8 @@ namespace rex
 
       void update()
       {
+        input::internal::update();
```
Dyronix, line 75: INFO: Is there a reason why input is stored within the internal namespace?

source/4_tools/console_app/console_app.sharpmake.cs
```diff
@@ -22,6 +22,13 @@ protected override void SetupConfigSettings(RexConfiguration conf, RexTarget tar
     conf.Options.Add(Options.Vc.Linker.SubSystem.Console);
   }
 
+  protected override void SetupOutputType(RexConfiguration conf, RexTarget target)
+  {
+    base.SetupOutputType(conf, target);
+
+    conf.Output = Configuration.OutputType.Exe;
```
Dyronix, line 29: INFO: Is this not the default output type of an application?

source/4_tools/console_app/src/console_app.cpp
```diff
@@ -1,21 +1,19 @@
...
 #include "rex_windows/input/input.h"
-#include "rex_engine/diagnostics/win/win_stacktrace.h"
-
-#include <Windows.h>
+#include "rex_engine/diagnostics/stacktrace.h"
 
 DEFINE_LOG_CATEGORY(LogConsoleApp, rex::LogVerbosity::Log);
 
-namespace rex
+namespace console_app
```
Dyronix, line 16: NIT: Maybe a more coherent name with rex and regina would be cool ^^,

source/1_engine/rex_engine/include/rex_engine/filesystem/path.h
```diff
@@ -216,6 +187,54 @@ namespace rex
       rsl::string_view root;
       rsl::string_view tail;
     };
+
+    // These functions are required to be implemented by platform specific code
```
Dyronix, line 191: Could you make the difference between platform code and generic code a bit more clear.
This line of comment is easily overlooked

Dyronix (changes requested):

source/1_engine/rex_engine/src/memory/memory_tracking.cpp
```diff
@@ -22,7 +22,9 @@
 #include "rex_std/bonus/time/win/win_timepoint.h"
 #include "rex_std/bonus/utility/yes_no.h"
 
-#include <vcruntime_new.h>
+// there's about 0.8MB allocation overhead of Window's process runtime
+// So during tracking, we need make sure we subtract this memory usage
+// from the app's usage, as there's nothing we can do about it
```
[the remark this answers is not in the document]
RisingLiberty: I tested this with debug code, I tested it on my machine, you can test it on yours if you like, I expect it to be similar.

source/2_platform/rex_windows/src/console_application.cpp
RisingLiberty: because I didn't want others to run the update, technically they're still able to do so, but it's in an internal namespace which implies that they shouldn't call it

source/4_tools/console_app/console_app.sharpmake.cs
RisingLiberty: yes, I just set it for readability

Dyronix: I think that would be implied from an "app" but ok

source/2_platform/rex_windows/src/console_application.cpp
Dyronix: But you are calling it?
Is there not a better approach that handle this, can't we PIMPL it?

source/1_engine/rex_engine/src/memory/memory_tracking.cpp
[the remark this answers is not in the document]
Dyronix: How are you going to "subtract" this from the memory tracker, just hardcode it?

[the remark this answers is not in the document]
RisingLiberty: probably just use it as "untagged" memory. the total memory usage of an app can be queried, you subtract all the memory you're aware, what's left over is untracked

source/2_platform/rex_windows/src/console_application.cpp
RisingLiberty: yes, I'm calling it from engine code. we can use pimp if you want to

Dyronix: If it does not over complicate things, please do I would use the internal namespace when there is no or no easy solution.

source/4_tools/console_app/src/console_app.cpp
RisingLiberty: I'll try to think of one

source/1_engine/rex_engine/include/rex_engine/filesystem/path.h
RisingLiberty: done

Dyronix (approved):

PR #242, opened 2023-12-26
Development/memory view

RisingLiberty:
Created rex::BlobView which is a view over a blob, but doesn't own the memory it contains


Dyronix (approved):

Dyronix:
I would have added conversion to a regular blob as well, like it's possible with a string_view -> string
But we can always add it when we need it.

PR #249, opened 2023-12-29
Development/using unicode

RisingLiberty:
Added unicode support for the filesystem and commandline as these are exposed to the user in some shape or form


_build/sharpmake/src/generate_settings.cs
```diff
@@ -167,7 +167,7 @@ public override string AsString(string key)
 
         foreach (string val in project_vals.Value)
         {
-          sb.AppendLine($"    {ElementType}{{ {val}, \"{project_vals.Key}\" }},");
+          sb.AppendLine($"    {ElementType}{{ {val}, L\"{project_vals.Key}\" }},");
```
Dyronix, line 170: Do we need this in the project generation files?
There is going to be a lot of overhead to keep track of where we need to introduce UNICODE and where we don't need to introduce it. 
I would try and limit it to one place ( maybe it can get converted to a UNICODE structure somewhere? )

config/tests/1_engine/rex_auto_test/code_generation.json
```diff
@@ -1,7 +1,7 @@
 {
     "CommandLineArgs": {
         "Content": [  
-            "\"AutoTest\", \"Specify the autotest to run\""
+            "L\"AutoTest\", L\"Specify the autotest to run\""
```
Dyronix, line 4: Same comment

source/1_engine/rex_engine/include/rex_engine/filesystem/vfs.h
```diff
@@ -6,5 +6,5 @@ namespace rex
 }
 
 #ifdef REX_PLATFORM_WINDOWS
-#include "rex_engine/platform/win/filesystem/vfs.h"
+#include "rex_engine/platform/win/filesystem/win_vfs.h"
```
Dyronix, line 9: Think I already commented on this before but just as a reminder, I think this has to be win/vfs instead of win_vfs

config/source/1_engine/rex_engine/code_generation.json
```diff
@@ -27,11 +27,11 @@
...
-      "\"Root\", \"The directory to initialize the vfs with. this is similar as setting the working root directory\"",
-      "\"Help\", \"This will list all command line arguments for the executable\"",
-      "\"LogLevel\", \"Specify log level per logger (eg. -LogLevel=Engine=trace,Renderer=info)\"",
-      "\"BreakOnBoot\", \"Break on boot so you can attach a debugger\"",
-      "\"AttachOnBoot\", \"Attach the debugger on boot\""
+      "L\"Root\", L\"The directory to initialize the vfs with. this is similar as setting the working root directory\"",
+      "L\"Help\", L\"This will list all command line arguments for the executable\"",
+      "L\"LogLevel\", L\"Specify log level per logger (eg. -LogLevel=Engine=trace,Renderer=info)\"",
+      "L\"BreakOnBoot\", L\"Break on boot so you can attach a debugger\"",
+      "L\"AttachOnBoot\", L\"Attach the debugger on boot\""
```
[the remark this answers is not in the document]
RisingLiberty: wdym? json doesn't know what the "L" means, this is so that the code generation, generates code (as in, uses wide chars), which makes it compile. otherwise I'd have to let the generation step take care of it and I don't think it belongs there

source/1_engine/rex_engine/include/rex_engine/filesystem/vfs.h
RisingLiberty: I commented in the redesign_vfs PR. I don't think it should be win/vfs

_build/sharpmake/src/generate_settings.cs
Dyronix: https://github.com/New-Creta/Rex/issues/281

config/tests/1_engine/rex_auto_test/code_generation.json
Dyronix: https://github.com/New-Creta/Rex/issues/281

PR #287, opened 2024-01-21
[development/better_memory_usage_logging] added memory logging at the first heap allocation, when we hit our entry point and after initialization

RisingLiberty:


source/1_engine/rex_engine/src/memory/global_allocator.cpp
```diff
@@ -1,12 +1,41 @@
...
+    bool log_mem_usage_before_tracking_impl()
+    {
+      log_mem_usage();
+      return true;
+    }
+
+    void log_mem_usage_before_tracking()
+    {
+      // Store result in static bool so we only log once
+      static bool _ = log_mem_usage_before_tracking_impl();
```
Dyronix, line 19: Can you give this local a proper name please.
I don't like this introduction of C# using _ as a temp variable name, please refrain from using it in C++.

source/1_engine/rex_engine/src/platform/win/diagnostics/win_debug.cpp
```diff
@@ -25,6 +25,12 @@ void rex::exit_confirm(int32 exitCode)
   }
 }
 
+void rex::output_debug_string(rsl::string_view str)
+{
+  OutputDebugStringA(str.data());
+  OutputDebugStringA("\n");
+}
```
Dyronix, line 32: There is a Sink within spdlog that uses "OutputDebugString" could you convert this to using this logger instead.
You can register a separate logger just for this use case.

Dyronix (changes requested):

source/1_engine/rex_engine/src/platform/win/diagnostics/win_debug.cpp
RisingLiberty: No I can't, this code is meant to be able to be called before the logger is initialized

Dyronix (dismissed):

Dyronix (approved):

